PR #6, opened 2026-05-27
feat(extension): make options page editable

rejifald:
## Summary
Ticket #5 of the v1 release punch list.

The options page displayed settings but had no interactive controls — to change any of them a user had to dig into `chrome.storage.sync` by hand. v1 needs a real settings UI.

### New sections
- **Language priority** — reorder (↑/↓), remove (×), and add from a dropdown. The dropdown's options are languages from the catalog not already in the list. A guard keeps at least one priority entry.
- **Blocked languages** — chips with a remove button + dropdown to block another. Can be emptied entirely.
- **Exempt sites** — text input + chip list. Domains are normalised (strip `http(s)://`, `www.`, paths, ports) and matched against a basic shape regex.
- **Page content** — checkbox bound to `settings.contentModification` (the flag added in ticket #4).

All mutations call `setSettings()` immediately. No save button, no draft state.

### Catalog
`SUPPORTED_LANGUAGES = ['uk', 'en', 'de', 'fr', 'es', 'it', 'pl', 'ru']` mirrors the "Preferred-language options" list in `apps/extension/STORE-LISTING.md`. Comment in the source flags the sync requirement when adding a new language.

### Design notes
- Side aside now also explains "blocked vs exempt" because the two adjacent sections looked interchangeable without it.
- Two small inline components (`IconButton`, `AddLanguagePicker`) — no new files, no new abstractions.

## Test plan
- [ ] Load unpacked, open options
- [ ] Reorder priority — confirm order persists across reload
- [ ] Remove the only priority entry — should be blocked (× disabled)
- [ ] Add a language to priority — should appear at the end
- [ ] Block/unblock a language — verify the popup status header reflects it
- [ ] Add `example.com` to exempt sites — visit `https://example.com`, confirm Movar takes no action
- [ ] Add `HTTPS://Example.com/path:8080` — should be stored as `example.com`
- [ ] Add a domain twice — should show "Already on the list"
- [ ] Toggle Page content on — confirm picker hiding / card blurring kicks in on a relevant site

### DIFF
--- a/apps/extension/src/entrypoints/options/App.tsx
+++ b/apps/extension/src/entrypoints/options/App.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useMemo, useState, type FormEvent } from 'react';
 import { browser } from 'wxt/browser';
 import {
   defaultSettings,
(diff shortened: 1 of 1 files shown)


rejifald:
Replaced by #12 — the original branch tripped the fallow metrics gate (App.tsx 326 lines / 6 cyclomatic). The replacement splits the component into per-section files so the audit passes. Same surface area.

PR #367, opened 2026-08-08
feat(youtube): cover the view-model DOM migration; harden SPA-click navigation guards

rejifald:
## What

Two related pieces of YouTube work, both driven by live inspection of today's youtube.com/m.youtube.com in a real browser:

1. **`@movar/page-content`: cover YouTube's polymer→wiz ("view-model") migration.** Three main surfaces stopped being extracted at all; the model now covers both markup generations.
2. **`apps/extension`: close the two residual holes in the aborted-SPA-click class** (#285) behind the "clicking a video blinks and never opens it" report.

## 1 — Page model: what the live survey found (2026-08-08, desktop + mobile-UA, logged out)

| Surface | Today | Model before this PR |
|---|---|---|
| /results video / channel / "latest from" cards | classic `ytd-*-renderer` | ✅ covered |
| **Watch-sidebar recommendations** | `yt-lockup-view-model` (`ytd-compact-video-renderer` is **gone**) | ❌ zero nodes on /watch |
| **Shorts shelf** (desktop AND mobile) | `grid-shelf-view-model` + `ytm-shorts-lockup-view-model` tiles (`ytd-reel-shelf-renderer` gone from /results; the `ytm-` prefix is served on desktop too) | ❌ missed |
| **Mix/playlist card on /results** | `yt-lockup-view-model` (`ytd-radio-renderer` gone) | ❌ missed |
| **m.youtube.com tiles** | still `ytm-video-with-context-renderer`, but **no `[id="video-title"]` anchor anymore** — the title's only durable anchor is its `<h3>` | ⚠️ matched the tile but serialized empty text → never hidden |

Design decisions (all per `docs/pitfalls.md` §1 and the google.ts selector discipline):

- **Anchors**: custom-element tag names + semantic ids/headings only — never wiz styling classes (`ytLockupViewModelHost`, …).
- **`fallbackTextSelectors: ['h3']`** on the classic shapes, consulted only when the classic anchors yield nothing — classic samples stay byte-identical (a classic tile's `<h3>` can carry badge chrome next to the title).
- **Lockups classify on their single `<h3>` title ALONE**, no fallback: the `yt-content-metadata-view-model` byline bakes the channel name together with UI-language view-count/date strings («139 тис. переглядів», «3 год тому») and the «Нове» badge — the same contamination class as Google's ad location extension. Fail open.
- **Per-element kind** (`kindFor`): collection-stack thumbnail → `playlist`; a shorts-less `grid-shelf-view-model` → generic `shelf`.
- **Outermost-wins dedup** (mirrors `extractGoogle`): a lockup inside its `ytd-rich-item-renderer` grid cell — and rich items inside a `ytd-rich-shelf-renderer[is-shorts]` — collapse onto the enclosing unit.

**Corpus**: two new real-capture fixture pairs (`results-shorts-shelf`, `watch-next-lockups`) pin the migrated markup, including the contaminating byline chrome, and encode **two documented false-keeps** (asserted `keep` so a classifier-side fix flips them loudly, same convention as `autotrans1`):

- a 4-ru/1-uk Shorts shelf whose pooled sample tips **uk** — the ru titles carry `#україна` hashtags (uk-specific letters) while lacking ы/э density;
- a marker-less Russian title («ПУТИН УЗНАЛ О ПРОВАЛАХ НА ФРОНТЕ…») that franc reads as uk (~0.23 margin; verified the channel name doesn't tip it).

Both are classifier-layer limitations (candidates: per-tile vote for shelf aggregates; better marker-less-text handling), out of scope for the DOM model.

**Live smoke test** (mirrored shapes run against the real pages): /results → 26 nodes (23 video / 1 channel / 2 shorts-shelf), zero empty samples, zero double-counts; /watch → sidebar lockups with clean h3-only text. Before this PR the same probe yields **0 nodes on /watch** and misses every Shorts shelf.

## 2 — The "can't navigate to a video, page blinks" report

**Primary cause is #285** (pre-commit `wxt:locationchange` → enforce rewrite `location.replace` aborting the click), fixed on `main` since 2026-07-22 — but it first reached a **published** release in **v1.5.2 (Aug 5) / v1.5.3 (Aug 7)**. v1.5.0 — the store version until Aug 5 — reproduces it deterministically, so reporters on a not-yet-updated extension still see it. Worth telling the reporter to update first.

Independently, I re-verified #285's fix against today's YouTube live (instrumented `navigate`-event vs commit vs mutation timings; simulated the full guard/gates/deferral ladder in-page on both www and m): the deferral holds on the normal path. The sweep for what's left found two real residual members of the class, both fixed here:

- **Drifted-`oldUrl` hole.** WXT's watcher advances its `lastUrl` on every dispatched `navigate` event — *including navigations that are canceled and never commit*. The next event's `oldUrl` then names a page never visited, and the `location === oldUrl` pre-commit tell fell through to the **synchronous** branch: `applyOnce` ran *inside* the click's navigate dispatch against the still-current URL (YouTube's page language is sync-detectable, so the ladder reaches `location.replace` without awaiting) — aborting the click whenever the current /results URL happened to be unguarded. The pre-commit tell is now measured against the **destination** (`location !== newUrl` → defer one macrotask), and the deferred reset compares against `priorHref` — the page we were really on at dispatch — never the drifted `oldUrl`.
- **Round-trip blink.** `applyRouteChange` cleared the loop guard on every pathname change, so clicking through to /watch wiped the `/results` mark — and the *return* to /results always found the guard empty, re-firing the hl/gl rewrite: a **full-reload blink on every results↔watch round trip** (and the widest unguarded window for the hole above). The route-change clear now skips enforce-mode hosts, mirroring the OK-page exception `applyOnceInner` has always carved out.

**Class sweep** (navigation writes racing an in-flight user navigation): `wxt:locationchange` has exactly one consumer (this handler, hardened); the redirect ladder, gate click, and `empty-results-retry` all fire from timer/tick contexts, which cannot interleave into a navigate dispatch — the URL commit is synchronous within it (measured live: `navigate` event at t≈26 ms after click, `location` already `/watch` in the same dispatch's `setTimeout 0`; `navigation.transition` is `null` throughout, so it is *not* usable as an in-flight signal). Popup crash-cards' `location.reload()` acts on the popup document, not the host page. `docs/pitfalls.md` §5 updated with both guards and the drift trap.

**Tests**: three regression tests pin the new behavior (drift deferral; enforce guard retention across the round trip; the #285 test updated for retention), plus the post-commit dispatch tests now stub a *committed* location, matching what they model.

## Verification

- `@movar/page-content`: typecheck ✓ lint ✓ **117 tests** ✓ (was 79)
- `apps/extension`: typecheck ✓ lint ✓ **1386 tests** ✓ (incl. the corpus over the two new fixtures)
- Extension build ✓ with chunk budgets (`models/youtube.js` well under 18 KB)
- `pnpm check:readme` ✓; coverage + metrics snapshot refreshed
- Pre-push: build ✓, offline e2e-fast 20/20 ✓ (one earlier run flaked under load-average ~40 from the coverage run — re-ran clean), metrics ✓

## Notes / follow-ups

- Classifier-side candidates (deliberately not in this PR): per-tile majority vote for shelf aggregates; marker-less-Cyrillic disambiguation. The two corpus false-keeps flip loudly when addressed.
- The entity "watch card" panel (`ytd-universal-watch-card-renderer`) on artist searches is still uncovered — semantics debatable (hiding the panel for an explicitly searched entity), left out.
- Live YouTube still strips `hl/gl` via replaceState (verified byte-equal restore), so the enforce dance and its guard remain load-bearing.


rejifald:
## Hot-path coverage matrix (all live-verified 2026-08-08, logged-out)

| Surface | Markup today | Extracted? |
|---|---|---|
| **Home feed** (`/`) | `ytd-rich-item-renderer` cells wrapping `yt-lockup-view-model` (23/32 cells), `ytd-rich-shelf-renderer[is-shorts]` Shorts shelves (2), `ytd-ad-slot-renderer` cells (3) | ✅ 26 video nodes + 2 shelf units; 31 nested candidates deduped; ad cells fail open (empty sample) |
| **Search** (`/results`) | `ytd-video-renderer`, `ytd-channel-renderer`, `ytd-grid-video-renderer` shelves, `grid-shelf-view-model` Shorts, `yt-lockup-view-model` Mix | ✅ 26 nodes on the test SERP, zero empty samples |
| **Watch** (`/watch`) | sidebar = `yt-lockup-view-model` | ✅ h3-only samples, byline chrome excluded |
| **Watch + queue** (`/watch?list=`) | queue = classic `ytd-playlist-panel-video-renderer`; sidebar lockups alongside | ✅ sidebar; queue **deliberately not extracted** (the visitor's own queue — documented stance) |
| **Channel Videos tab** | rich-grid cells wrapping lockups (30/30, no `#video-title`) | ✅ 30 nodes, all inner lockups deduped |
| **Channel Playlists tab** | standalone collection-thumbnail lockups (30) | ✅ kind `playlist` via `kindFor` |
| **Playlist page** (`/playlist`) | rows are `yt-lockup-view-model` — `ytd-playlist-video-renderer` is **gone** here | ✅ via the lockup shape |
| **Shorts player** (`/shorts/ID`) | `ytd-reel-video-renderer` + overlay view-models, no card lists | ⛔ deliberately out of scope (immersive player, same stance as the /watch player) — documented |
| **m.youtube.com** search + watch-next | `ytm-video-with-context-renderer` (h3-only titles), `grid-shelf-view-model` Shorts | ✅ via h3 fallback |
| m.youtube.com home | not observable in this session | covered by construction (`ytm-rich-item-renderer` + grid-shelf + h3 fallback) |

Follow-up candidates (not gaps in this PR's scope): YouTube ad cells could get a Google-style `ad` kind instead of failing open; the artist "watch card" panel remains uncovered by choice.

The extractor header now records this survey and the three deliberate non-goals, so future audits can tell stance from gap (`dab621b`).

rejifald:
Bughunt fixes from the dogfood run are now on the branch:

- `ca44f27` **feat(page-content): model the YouTube channel Posts surface** (#371) — 'post' shape (`#content-text` alone, chrome excluded, fail-open), 5 unit tests, 2 real-capture fixtures (classifier pre-verified), live shape-mirror 10/10 on /@tsn/posts.
- `940e6f0` **fix(sites): allowlist udm=web and scope model chunks to parseable hosts** (#370, #372) — Google's own Web tab gets hl/lr again; model chunks no longer provision on music/studio/kids/news/scholar/translate.

Verified end-to-end: extension 1402 / page-content 122 / host-match 67 tests, lint+typecheck clean, build budgets hold, coverage snapshot refreshed, pre-push (build + e2e-fast 20/20 + metrics) green. Issues #370–#372 auto-close on merge; the YouTube-Music adoption decision is flagged on #372 to survive it.

PR #386, opened 2026-08-11
ci(safari): automate Safari releases — the CI 401 was a malformed issuer ID, not headless provisioning

rejifald:
Safari has never shipped from CI. `release-safari` failed at the archive step for v1.2.0, v1.3.0 and v1.4.3, and the cause was recorded in the docs as *"headless automatic signing can't mint provisioning profiles on the runner"*.

**That diagnosis was wrong.** `APPLE_ASC_ISSUER_ID` simply wasn't a UUID — a Key ID or Team ID had been pasted into it — so every App Store Connect call 401'd, including the one Xcode makes during `archive`. Apple's own tooling says so without any inference:

```
Error: The value '***' is invalid for '--issuer <issuer>': *** must be a valid UUID
```

Three releases of "we can't automate Safari" were one malformed secret.

## Credential verdict

Built a read-only audit rather than guessing, then fixed what it found. Every other secret was healthy:

| Secret | Verdict |
| --- | --- |
| `APPLE_ASC_ISSUER_ID` | ❌ → fixed (was not a UUID) |
| `APPLE_ASC_KEY_ID` + `APPLE_ASC_API_KEY_P8` | ✅ authenticates, **Admin**, Cert/ID/Profiles access, notarytool OK |
| `APPLE_TEAM_ID` | ✅ |
| `APPLE_DIST_CERT_*` | ✅ valid to 2027-06-23, team `RQSR4UU3VB` |
| `APPLE_DEVELOPER_ID_CERT_*` | ✅ valid to 2027-02-01 |
| **`APPLE_INSTALLER_CERT_*`** | ➕ **created during this PR** — never existed |

Nothing needed *regenerating*. One certificate needed *creating*.

## What this PR adds

**`safari-credentials-audit.yml` + `scripts/apple-credentials-audit.mjs`** — read-only, `workflow_dispatch`, every Apple call a `GET`. Mints ASC JWTs itself and probes `/v1/apps`, `/v1/certificates`, `/v1/users`, `/v1/bundleIds`, `/v1/profiles`; inspects each `.p12` the way the release job does and cross-checks its serial against what Apple still lists (a revoked cert imports just as cleanly as a live one); probes notarytool separately. It distinguishes a bad triple from an under-privileged key from an Individual-vs-Team key, and names the *shape* of a malformed value without printing it.

**Manual signing** (`scripts/apple-provisioning.mjs`) replacing automatic signing. Two independent reasons, both measured:

- `-exportArchive` with `signingStyle: automatic` means *cloud signing*, and it fails with `Cloud signing permission error` **even though the key is Admin** with full Cert/ID/Profiles access. No `DISTRIBUTION_MANAGED` cert has ever been issued to this team. Adding the omitted `teamID` changed nothing — tested and reverted rather than left in as cargo cult. Organizer succeeds locally precisely because it *doesn't* cloud-sign.
- Automatic archiving minted a throwaway `Apple Development` cert **every run** (the account now shows one as `DEVELOPMENT — Created via API`), against Apple's per-team cap.

The script fetch-or-creates profiles, binds each to the certificate whose serial is in *this* keychain, installs them, and generates the export plists alongside — so names can't drift. Profile names derive from the bundle ID, so one `PROVISIONING_PROFILE_SPECIFIER=movar $(PRODUCT_BUNDLE_IDENTIFIER) …` covers app and extension (xcodebuild expands it per target — the only way to vary signing per target from the CLI). `project.pbxproj` is untouched, so the local Organizer flow is unaffected.

**Xcode floor instead of an exact pin**, in all three macOS workflows. The exact `Xcode_16.4` pin was chosen for reproducibility; it's what a moving external floor costs you — App Store Connect rejected a correctly signed `.ipa` with `SDK version issue … must be built with the iOS 26 SDK or later`. Runners have 26.3.

**Docs corrected.** Both docs asserted the wrong root cause for three releases. Replaced with what was measured, plus why manual signing and the Xcode floor are load-bearing so neither gets "simplified" back. The two committed `exportOptions/*.plist` are **deleted** — they encoded the cloud-signing approach that cannot work here.

**Mac Installer Distribution certificate.** A macOS App Store artifact is a `.pkg`, signed by a *different* certificate from the `.app` inside it. None existed in the account or the keychain. Created and stored as `APPLE_INSTALLER_CERT_*` during this PR. The identity's name is resolved from the keychain at runtime rather than hardcoded — Xcode's error calls it `Mac Installer Distribution` while the CN is `3rd Party Mac Developer Installer`, and either string would be a guess that breaks on the next rename.

**altool exits 0 even when it fails.** A run went green while its own output said `VERIFY FAILED with 1 error` and App Store Connect had rejected the package. Every altool call was decorative — and `release-safari` uses `--upload-app`, so a failed *submission* would have been swallowed just as quietly. Now parses altool's JSON report and gates on `product-errors`, treating unparseable output as failure.

**Build numbers had to move to timestamps.** The Mac App Store requires each upload's `CFBundleVersion` to exceed the last accepted one *across marketing versions*. CI passed `github.run_number` — monotonic, but `9` — against a previously accepted `1785959230` and was rejected. The manual submissions always used `date +%s`; once timestamps are in an app's history the schemes can't be mixed.

## Proof

`safari-signing-rehearsal.yml` (upload-free) is **green for both platforms**:

```
Installer identity: 3rd Party Mac Developer Installer: … (***)
** ARCHIVE SUCCEEDED **   ×2   (iOS, macOS)
** EXPORT SUCCEEDED **    ×2   (.ipa, .pkg)
VERIFY SUCCEEDED with no errors
  "success-message" : "No errors validating archive at '…/Movar.ipa'"
VERIFY SUCCEEDED with no errors
  "success-message" : "No errors validating archive at '…/Movar.pkg'"
```

Both packages accepted by App Store Connect's validator. Safari can ship from CI like every other store; the Organizer walkthrough stays in the docs as a fallback.

## The submission itself is automated too

`altool --upload-app` only *delivers a build*. Creating the version, attaching the build once Apple finishes processing it, writing "What's New" per localization, answering export compliance and submitting were hand-work in the App Store Connect UI — this doc said so. The **Review Submissions API** covers all of it, so `scripts/apple-submit.mjs` + `safari-submit.yml` finish the job.

Three levels, mirroring release.yml's "reversible first, irreversible last":

| Mode | Does | Reversible? |
| --- | --- | --- |
| `plan` | reads only; prints exactly what it would do | n/a — no approval needed |
| `prepare` | create/reuse version, attach build, set notes, export compliance | yes, all editable in ASC; no reviewer sees it |
| `submit` | also submits for review | **no** |

Idempotent at each level: an editable version is reused rather than duplicated, and a version already `WAITING_FOR_REVIEW` or `READY_FOR_SALE` is reported and left alone.

Release notes come from the existing `WHATS-NEW.md` — human prose interleaved with fenced note blocks, of which only the fences ship. That parser has [its own test](scripts/lib/whats-new.test.mjs) asserted against the **live** file, because Apple rejects a version whose localizations have no "What's New": a parser that silently returned nothing would surface as a metadata rejection at the submission step, far from the mistake. A locale with no note is a hard error naming the locale.

Separate workflow from `release-safari` deliberately — Apple's processing takes minutes to tens of minutes and waiting on a 10x-priced macOS runner is waste; it's pure HTTP, so it runs on Linux, and it's re-runnable without rebuilding or re-uploading.

**Verified against the real API** (read-only plan):

```
app: Мовар (6779282071)
release notes: uk, en
▸ IOS — 1.6.2      build 1786451798 is processed (VALID)
▸ MAC_OS — 1.6.2   build 1786451798 is processed (VALID)
```

## Note

`release-safari` still parks on the `production` environment gate, and so does `safari-submit`'s `apply` job. Merging this changes nothing until a release is published and approved.

The first real run already happened from `release/safari-v1.6.2` (a separate, deliberately unmergeable branch): both 1.6.2 builds are uploaded to App Store Connect and the notarized `.dmg` was produced.


rejifald:
## Audit result — one secret is wrong, and it isn't a cert

Run [31478935193](https://github.com/rejifald/movar/actions/runs/31478935193).

| Secret | | Finding |
| --- | --- | --- |
| `APPLE_TEAM_ID` | ✅ | Well-formed 10-char Team ID. |
| `APPLE_DIST_CERT_P12_BASE64` / `_PASSWORD` | ✅ | Imports (`1 identity imported`). `Apple Distribution: … (RQSR4UU3VB)`, expires **2027-06-23**, serial `4D2DB1DCA2AAA49DA00ECAEA499C71DD`. |
| `APPLE_DEVELOPER_ID_CERT_P12_BASE64` / `_PASSWORD` | ✅ | Imports. `Developer ID Application: … (RQSR4UU3VB)`, expires **2027-02-01**, serial `3B1BE817A166EBF5`. |
| `APPLE_ASC_API_KEY_P8` | ✅ | Decodes to a valid EC prime256v1 key. Not yet provable against Apple — see below. |
| **`APPLE_ASC_ISSUER_ID`** | ❌ | **Not a UUID.** Shape: 10 chars, uppercase alphanumeric — i.e. a **Key ID or Team ID** was pasted into the issuer slot. |
| `APPLE_ASC_KEY_ID` | ❔ | Unverifiable while the issuer is malformed; every call 401s before the key id is reached. |

Both cert serials match the identities in the maintainer's login keychain exactly, so the `.p12` secrets hold the certs that were verified locally as valid into 2027. **No certificate needs regenerating.**

### The 401 is explained

Apple's own tooling confirms it without any inference on our part — `notarytool`
refuses the value before it even opens a connection:

```
Error: The value '***' is invalid for '--issuer <issuer>': *** must be a valid UUID
```

A malformed issuer means every App Store Connect JWT is unsignable-as-valid, so
every call 401s — including the `listTeams.action` one that Xcode's automatic
signing makes during `xcodebuild archive`. That is the `** ARCHIVE FAILED **`
this whole thread has been about, and it has been the same single character-level
mistake since v1.2.0.

### Fix

Copy the **Issuer ID** (the UUID above the key list at App Store Connect →
Users and Access → Integrations → App Store Connect API) and:

```bash
gh secret set APPLE_ASC_ISSUER_ID
```

Nothing else needs rotating on current evidence.

### What the audit still can't rule out

Two things stay unknown until the issuer is valid and the audit reruns:

1. **The key's role.** `docs/safari-deploy.md` recommends **App Manager**, which does *not* grant Certificates, Identifiers & Profiles access — the exact access automatic signing needs. The audit now probes `/v1/certificates` and will say so directly (200 → fine, 403 → regenerate as Admin). If that lands 403, the doc's recommendation is itself a bug worth fixing.
2. **The archive's development-signing round-trip.** `xcodebuild archive` under automatic signing resolves an *iOS App Development* profile before `-exportArchive` re-signs for distribution — note the original error said "couldn't find any **iOS App Development** provisioning profiles". The CI keychain holds only Distribution + Developer ID, so Xcode would mint a development cert on each run, against a per-account cert limit. If that becomes the next failure, the durable fix is manual signing with pre-created profiles, which removes the Developer-portal round-trip from the archive step entirely.

The temporary branch `push:` trigger stays until the issuer is fixed and the audit is green; it comes off before merge.

Issue #401, opened 2026-08-12
be/bg conceal as ru under the shipped candidate set: an accepted #103 tradeoff asserted otherwise by two tests, a fixture, and the store-policy doc


rejifald:
Movar conceals Belarusian and Bulgarian as Russian when the content filter is on.
That much is a **known, deliberately accepted** tradeoff from #103 — this issue is
not asking to reverse it. The defect is that the acceptance was never recorded
where anyone would find it, and three surfaces now actively assert the opposite:
two tests, a fixture, and the doc written for store reviewers.

Found during issue triage; verified against `main` @ `ab2451a`.

## What ships

`apps/extension/src/lib/content-modification.ts:168` derives the classifier's
candidate set as `priority ∪ blocked`, gated on `hasProfile`:

```ts
const enabled = new Set(settings.priority.filter(hasProfile));
const candidateCodes = [...new Set([...settings.priority, ...settings.blocked])].filter(hasProfile);
```

With the default `priority = ['uk','en']` and derived `blocked = ['ru']`, that is
**`{uk, en, ru}`**. `PROFILED_CODES` ships `{uk, ru, be, bg, en}`
(`packages/lang-detect/src/profile-codes.ts:11`), so the `be` and `bg` profiles
exist but are never candidates for a default user.

Rung 1 is candidate-set-relative — a letter counts only if it appears in exactly
one *candidate's* alphabet — so dropping `be`/`bg` from the set does not merely
lose their signal, it **reassigns their letters to `ru`**:

| letter | in `be`? | distinctive under `{uk,en,ru}` | distinctive under `{uk,en,ru,be,bg}` |
| --- | --- | --- | --- |
| `ў` | yes (uniquely be) | **nothing** — in no candidate alphabet | `be` |
| `ы` `ё` `э` | yes | **`ru`** | inert (shared `ru`/`be`) |
| `і` | yes | `uk` | inert (shared `uk`/`be`) |
| `ъ` | no (bg) | **`ru`** | inert (shared `ru`/`bg`) |

Belarusian's strongest signal contributes zero, and the letters it shares with
Russian become confident Russian evidence.

## The repo's own fellow-victim fixtures invert under the shipped set

Taking the two fixture texts from `packages/lang-detect/test/fixtures.ts` verbatim
and tallying rung-1 distinctive characters against the shipped profile alphabets:

[...]

Issue #103, opened 2026-06-13
Detection: stop concealing fellow-victim Cyrillic languages (be/bg) as Russian


rejifald:
## Current state

- The content filter's candidate set is `candidateCodes = [...new Set([...settings.priority, ...settings.blocked])]` (`apps/extension/src/lib/content-modification.ts:145`). `enabled = new Set(settings.priority)` (line 144). Default `blocked` is `['ru']` (the only unconditionally-locked imposer per `docs/per-snippet-language-detection.md` decision 6/§"Conceal predicate"). So for a default UA user the candidate set is effectively `{uk, ru}` (plus `en` if priorited) — **`be` and `bg` are never candidates unless the user manually adds them to priority/blocked.**
- `classifyBySnippet` (`packages/lang-detect/src/classify.ts:240`) is candidate-set-relative: a letter is distinctive for a language only if it appears in exactly one *candidate's* alphabet (`distinctiveChars`, line 125; `tally`, line 159). With only `{uk, ru}` in scope and `ru.alphabet` containing `ъ` (`packages/lang-detect/src/profiles.ts:80`), the Bulgarian hard-sign `ъ` reads as ru-distinctive and pulls Bulgarian text to a confident `ru` verdict at rung 1.
- `PROFILES` ships only `uk`, `ru`, `be`, `en` (`packages/lang-detect/src/profiles.ts:193`). **There is no `bg` profile** — Bulgarian has no alphabet/word data, so even if it were imposed as a candidate it could only ever be detected by elimination, and today its signature letter `ъ` is owned by `ru`.
- `tokenize` (`classify.ts:150-152`) lowercases and splits on `/\p{L}+/gu`, which **drops the intra-word apostrophe** (`під'їзд`, `комп'ютер`, `сім'я`). The legacy `detectCyrillicLanguage` path treated the apostrophe as a uk/be signal — `packages/lang-detect/src/index.test.ts:19` asserts `isRussian('під'їзд') === false` — but the shipped `classifyBySnippet` has no such signal. The ADR parks this as an Open Question ("Orthographic-mark evidence", `docs/per-snippet-language-detection.md:228`); it was never landed.
- `dominantScript` (`classify.ts:75-84`) counts **raw characters** in a `for...of` loop and returns the script with the higher count. `scopeCandidates` (line 97) then restricts candidates to that script, so a card whose script vote flips to Latin scopes to `{en}` only and can no longer be detected as Russian.
[...]

Issue #58, opened 2026-06-05
Diagnostics: override → calibration back-feed loop (deferred / vNext)


rejifald:
**Milestone:** vNext (backlog)
**Status:** Planned, deferred — build **after** the card-scoped diagnostics view (v1.2.0 milestone) lands.

Close the calibration loop from the diagnostics UI: let a maintainer **override** a per-item language verdict and **back-feed** the correction into the classifier's thresholds. The calibration harness already exists (`packages/lang-detect/scripts/calibrate.mts`: corpus → `runSweep` → recommended `(lengthFloor, hideMargin)` vs current `(24, 0.22)`); this wires a human correction into it.

### Intended design (decided 2026-06-05: "export labeled samples → calibrate")
1. **Override UI** — in the card-scoped diagnostics panel, a maintainer corrects the language verdict (the gold label) on an item the classifier got wrong.
2. **Accumulate locally** — corrections collect on-device (in-memory + `storage.local`); never networked (on-device invariant, `docs/per-snippet-language-detection.md` decision 10).
3. **Export** — an action emits `CalibrationSample[]` (`{ id, text, gold, source: 'diagnostics', note }`) — the exact shape `calibrate.mts` already consumes (`packages/lang-detect/src/calibrate-types.ts`).
4. **Calibrate** — feed the export into the residual corpus (`packages/lang-detect/test/calibration-corpus.ts`) and run `pnpm --filter @movar/lang-detect calibrate` → sweep → recommendation.
5. **Commit** — update `packages/lang-detect/src/thresholds.ts` if the sweep improves leak / false-hide (block-only asymmetry: `maxFalseHide = 0`).
6. **(Optional) structure override** — when the extractor missed or mis-typed a card, capture it as an extractor test fixture (separate from the language corpus).

### Why deferred
The card-scoped view + the extractor promotion are prerequisites; the loop is additive and is best tuned once there's real card-scoped observation data. Keeps v1.2.0 shippable sooner.

### Definition of Done (when scheduled)
- [ ] Per-item override control in the diagnostics panel.
- [ ] On-device correction store (no network).
- [ ] Export → `CalibrationSample[]` consumable by `calibrate.mts` with no format massaging.
- [ ] Documented round-trip (override → export → corpus → calibrate → thresholds) in `apps/diagnostics/README.md` and/or `docs/rung3-threshold-calibration-harness.md`.
- [ ] Boundary invariant still holds (dev extension only; product untouched).

### References
[...]

PR #402, opened 2026-08-12
docs(detection): say plainly that bg conceals as ru — and that be no longer does

rejifald:
Documentation, naming, and one characterization test — **no behaviour change**, no `src/` runtime change.

Movar conceals **Bulgarian** as Russian once content filtering is on. **Belarusian is
held back** by the accounting veto. Four surfaces asserted neither, and one of them is
the doc written for store reviewers.

## What this branch originally said, and why it changed

As cut on 2026-08-12, this branch recorded that Belarusian *and* Bulgarian alike reach
a confident `ru` verdict under the shipped candidate set. That was true when measured.
Six days later [#523](https://github.com/rejifald/movar/pull/523) bumped langtell to
0.6.1, which **withdraws a winner whose own alphabet cannot account for 2% of the
text's letters** — and that splits the two fellow victims apart. Re-measured against
current `main` before landing:

| | letters `ru` cannot account for | verdict under shipped `{uk, en, ru}` | outcome |
| --- | --- | --- | --- |
| `be` | `і` `ў` (+ apostrophe marks) | `unknown`, margin 0 | **kept** |
| `bg` | *(none)* | `ru`, rung 1, margin 1–2 | **concealed** |

The split is structural, not incidental: it follows from alphabet containment. Belarusian
has letters Russian does not, so the veto has something to fire on; every Bulgarian letter
is in Russian's alphabet, so it can never fire on Bulgarian's behalf. `minHideMargin(1) = 1`,
so bg's margin clears the conceal gate.

## Why the letters flip in the first place

Candidates are `priority ∪ blocked`, so a default profile classifies against `{uk, en, ru}`
and the shipped `be`/`bg` profiles are never in scope. Because distinctiveness is
candidate-relative, dropping them does not merely lose their signal — it **reassigns their
letters**:

| letter | distinctive under `{uk,en,ru}` (shipped) | under `{uk,en,ru,be,bg}` (tests) |
| --- | --- | --- |
| `ў` | nothing — in no candidate alphabet | `be` |
| `ы` `ё` `э` | **`ru`** | inert (shared `ru`/`be`) |
| `ъ` | **`ru`** | inert (shared `ru`/`bg`) |

That reassignment is what elects `ru`. The veto is the separate, later step that can take
the election away again — and only Belarusian gives it grounds to.

## The tradeoff is still the #103 one

Re-adding the be/bg overlay stays a non-goal. #103 landed an always-on overlay and reverted
it in the same PR: adversarial e2e showed it diluting franc's Russian margin on a real RU
title (0.222 → 0.194, under the 0.22 conceal gate), making Movar miss a Russian card.
Missing Russian outranks a fellow-victim false-hide. That makes the remaining Bulgarian gap
a calibration question (#58's harness), not a candidate-set question.

## Changes

- **`store-policy-stance.md`** — claimed flatly that Belarusian "stays visible". Now states
  the safeguard that actually protects it, and names Bulgarian as the language that still
  needs the caveat. This is the trust-critical one: the doc exists to be read by store
  reviewers, and the original wording would have told them Belarusian can be concealed —
  understating a protection that exists while misnaming the affected language.
- **`per-snippet-language-detection.md`** — the limitation is written down in the
  conceal-predicate section with both measurements, the containment reason they differ, why
  re-adding the overlay is not the fix, and #58's calibration harness as the path forward.
- **`classify-shipped-candidates.test.ts`** (new) — pins the asymmetry: four `be` texts
  abstain, two `bg` texts conceal above the gate, genuine Russian still hides, and the
  alphabet-containment premise itself is asserted. This is the characterization test the
  original scope deferred; it is worth landing now precisely *because* the docs assert the
  behaviour, so they have something to drift against.
- **`classify-franc.test.ts`** — `DEFAULT_CANDIDATES`, documented as
  `priority ∪ blocked ∪ imposed be/bg overlay` (an overlay that no longer exists) →
  `FULL_ROSTER`, documented as deliberately *not* the shipped set, and cross-linked to the
  new shipped-path suite.
- **`fixtures.ts`** — `be-not-concealed-as-ru-default-candidates` is named for a set the
  product never builds, and pins the engine roster + legacy heuristic rather than the
  snippet classifier. Renamed and rescoped; the section header now states the boundary and
  the be/bg divergence under it.
- **`glossary.md`** — `bg` "distinguished by the detector" → when it is a candidate.
- Also corrects a stale Open Question in the ADR: orthographic-mark evidence is listed as
  "parked 2026-06-03", but the `marks` field shipped with #103 and is live in the profiles
  (`uk`/`be` → `'’ʼ`), now sourced from langtell and read by `distinctive.ts`.
- **`readme-metrics.snapshot.json`** — refreshed. Note this absorbs **354 lines of
  pre-existing `loc` drift on `main`** (65117 committed vs 65471 live). It was inside the
  500-line tolerance, so it was not failing the gate; only 13 of the +367 are this branch's
  (`test/fixtures.ts` is counted as source — `isTest` matches `*.test.ts`, not `fixtures.ts`).

## Verification

Re-verified against current `main`, not the August run: `@movar/lang-detect` 293/293, lint
and typecheck clean, coverage recomputed across 21 projects, `metrics-gate` green with `loc`
matching a live scan exactly. No changeset — docs, test naming, and test-only additions.

## Scope

Refs #401, items 2–4, plus item 1 (the characterization test) now that the behaviour it
would pin is the thing the docs claim. Left open there: the #58 cross-link.


rejifald:
## Review — REQUEST_CHANGES

The technical substance is accurate and was independently verified by running the classifier against the shipped `langtell@0.6.0` profiles: with the default candidate set `{uk, en, ru}`, `ў` is inert and `ы/ё/э/ъ` are ru-distinctive, both fellow-victim fixtures return `{language:'ru', margin:2, rung:1}`, and `minHideMargin(1) === 1` — so margin 2 clears the conceal gate and the doc's "confident rung-1 `ru` verdict … concealed" claim is literally true. The rename is honest: `FULL_ROSTER` is the same array, no assertion was removed or loosened, and the fixture keeps `expectedEngineLanguage: 'be'`. Prettier clean on all five files.

What blocks it is the docs half — four accuracy defects, two newly introduced here.

### Should-fix

**`docs/per-snippet-language-detection.md:230`.** "pinned by the `uk-apostrophe-pidyizd` fixture" is false. That fixture is consumed only by `fixtures.test.ts` (`detectCyrillicLanguage`) and `engines/franc.test.ts` (`francEngine`) — neither calls `classifyBySnippet`. Stripping `marks` from the `uk` profile leaves the verdict `uk` (margin 3→2) and nothing fails. The test that did pin it was deleted in #157. This also contradicts the boundary this same PR writes into `packages/lang-detect/test/fixtures.ts:139-145`.

**`docs/per-snippet-language-detection.md:27`.** Still lists the orthographic-`marks` polish under "**Still open:**" while line 230 now says it landed in #103. The PR half-corrected a stale statement and left the document contradicting itself.

**`docs/store-policy-stance.md:6` and `:54`.** The frontmatter `summary` and the §3 bold heading still make the flat claim ("fellow Cyrillic / 'victim' languages stay visible") that this PR's own body now qualifies — the two places a store reviewer skimming reads first.

**`apps/extension/store-assets/apple/REVIEW-NOTES.md:69` and `deployment-checklist.md:13`.** Both carry the pre-PR wording verbatim, and REVIEW-NOTES.md declares itself a mirror of `store-policy-stance.md`. After this merges the mirror is broken on the copy actually submitted to Apple — which is the reason to fix this before landing rather than after.

### Nits
- `packages/lang-detect/src/classify-franc.test.ts:11` — "`be`/`bg` ship profiles but are never candidates" is unqualified; `content-modification.ts:161` notes a `be` profile does become a candidate if a user puts `be` in `priority`.
- `docs/store-policy-stance.md:67` — "the languages that user has chosen" is missing an article; "a detection limitation being worked" wants "worked on".
- `classify-franc.test.ts:88` (pre-existing) — the test is named "…franc-ranks ru" but the verdict resolves at rung 2b under both candidate sets; franc never fires.

*Automated review.*

PR #423, opened 2026-08-14
feat(marketing): add the «Для української» directory, unlisted for now

rejifald:
A bilingual public directory of initiatives working for the Ukrainian language, at `/for-ukrainian` and `/uk/for-ukrainian`, with 20 verified entries across seven spheres. Plus the design record behind it.

Three commits: the plan, the build, and a correction where main overtook the plan mid-write.

## The page is deliberately not findable yet

That takes three separate mechanisms, and only having all three makes it true:

- `noindex` on both locales, via a new `BaseLayout` prop
- both routes excluded from the sitemap
- no inbound link anywhere

Not linking a page does none of this work on its own — a static build ships every page and the CDN serves it to anyone who asks. The plan carries a go-public checklist for reversing all three together, plus the two pieces that must land with them (the footer link's href helper, and the middleware's `UK_COUNTERPART` entry).

## What the directory is

Listing is **unconditional**. Nothing is asked of a listed project — no link back, no mention, no payment — and the page says so outright, because a message from a browser extension to a project it has listed reads as the opening of a pitch unless it doesn't.

Nothing is ranked, scored or tiered. There is no schema field for it. Entries sort alphabetically by the page locale's collator, which gives the two locales genuinely different orders, and the page explains that so the result doesn't read as arbitrary.

Individuals may be listed, but only with consent recorded before publication. This is enforced by the schema rather than by remembering: an entry naming a person cannot ship without a `consent` record, and only such an entry may carry one.

## Two rejected requirements, recorded with reasoning

A reciprocal-link condition and a Movar Audit conformance gate. Both were considered and dropped; the reasoning is in the plan so neither gets re-proposed from first principles.

The third commit strikes one of the four grounds recorded against the audit gate — that it would block on unbuilt software — because `@movar/audit` shipped on main while the plan was being written (#406, #424). Struck rather than deleted, with a note that the rejection stands on the other three. Quietly removing a falsified argument would leave the record looking stronger than it was.

## Guards for accepting submissions

`apps/marketing/src/content/projects/` is the first path in this repo that expects pull requests from outside it, so:

- a CODEOWNERS entry for that path
- `directory-submission-guard.yml`, which fails a **fork** PR that edits directory entries and anything else. Fork-scoped on purpose: in-repo branches already require write access, and maintainer PRs legitimately change entries and page code together — as this one does.

The other half of that mitigation is a repository setting, documented in the workflow: Settings → Actions → General → require approval for external contributors. **Not yet enabled.**

## Verification

Lint, typecheck, format, the full pre-commit suite, and `nx run marketing:audit` all pass. Four Linux visual baselines added for the new route; no existing baselines disturbed. Checked in-browser in both locales, light and dark.

## Not included

The outreach templates, and the entries queued behind outreach — three individuals awaiting consent and one dormant project awaiting a reply. None of them appear on the page until an answer comes back.


rejifald:
## Review — REQUEST_CHANGES

The headline claim holds. `noindex, nofollow` is present on both `dist/for-ukrainian/index.html` and `dist/uk/for-ukrainian/index.html`; the built `sitemap-0.xml` has 18 `<loc>` entries and zero matching `for-ukrainian`; grepping all 22 built HTML files finds `for-ukrainian` only in the two pages themselves. Not over-hidden either — both routes serve and `check:locale-redirects` passes. All 21 external URLs return 200 with correct domains, all render `target="_blank" rel="noopener noreferrer"`. `astro check` 0/0/0, `eslint` clean, `check:readme` passes, audit over `dist/` exits 0 and is byte-identical to a run with the two new pages removed. The Ukrainian copy reads as natively written and uses U+02BC consistently with the existing 26 occurrences.

### Blocking

**`apps/marketing/src/content.config.ts:83` (and `:95`) — security.** `z.string().url()` accepts any scheme. A probe entry with `url: 'javascript:alert(document.domain)'` builds successfully and emits `href="javascript:alert(document.domain)"` into `dist/for-ukrainian/index.html`. There is no CSP (`apps/marketing/public/` has no `_headers`), so it executes on click. This is the one path the feature designs for outside contributors, and the schema is already the PR's chosen enforcement mechanism — it enforces consent that way — so a scheme check belongs beside it rather than relying on a reviewer eyeballing 20 lookalike frontmatter diffs.

**`.github/workflows/directory-submission-guard.yml:57`.** `git diff --name-only "$BASE_SHA" "$HEAD_SHA"` is a two-dot diff, but `pull_request.base.sha` is the base-branch tip at event time, not the merge base. Once main moves ahead of the fork's branch point — 5 commits landed while this PR was open — the diff also reports every file main changed, and a one-file submission fails with "These files are outside it: pnpm-lock.yaml …". That is an accusatory false positive aimed at exactly the contributors this feature courts. `fetch-depth: 0` is already set, so `"$BASE_SHA"..."$HEAD_SHA"` fixes it.

**`content.config.ts:82` + `docs/articles/for-ukrainian.plan.md:356`.** Both say the `url` is "Checked by `pnpm check:projects`", and the plan promises "a CI link check turns rot into a build failure". No such script exists — no `check:projects` in `package.json`, no such file in `scripts/`. Written in the present tense it reads as shipped, so the next person won't add it.

**`apps/marketing/astro.config.mjs:13`.** The sitemap exclusion is an exact string match against a URL shape Astro controls. Flipping `build.format` to `'file'` or setting `trailingSlash: 'never'` makes `UNLISTED` match nothing, nothing fails, and the page starts being advertised to crawlers. Nothing in the repo asserts sitemap contents.

### Nits
- `apps/marketing/AGENTS.md` untouched — its route table and content-collection section still describe only `blog` and the nine existing page pairs.
- `apps/e2e/src/marketing/marketing.visual.spec.ts:13` docblock still says "8 pages … = 32 baselines"; there are now 9 full pairs.
- `src/i18n.ts:619`/`:1459` — `forUkrainian` is first in both locale literals but last in the `Strings` interface.
- uk `activeAtLabel` "Активні тут github.com" is plural where the one entry using it is feminine singular; reviewer was not fully certain.
- Commit `72746e5` says "six spheres"; it is seven. The PR body is correct.

*Automated review. Findings were verified by building the PR and probing the built output, not by reading alone.*

rejifald:
Pushed `8cf0aaa` fixing the blocking security finding.

`url` and `activeAt` now share a `linkUrl` validator that keeps `.url()`'s parse check and adds an `http(s)`-only scheme check. It sits beside the consent rule and follows the same reasoning — a rule a human has to remember is a rule that eventually ships broken — and it adds no test file because this app has no vitest config; the schema *is* the enforcement, checked at build time, exactly as the consent refine is.

Verified on this branch:
- probe entry with `url: 'javascript:alert(document.domain)'` → build fails with `url: must be an http or https URL`
- probe removed → 22 pages built, `astro check` 0 errors / 0 warnings / 0 hints, eslint and prettier clean

The validator stays silent when `new URL()` throws, so a malformed URL is reported once by `.url()` rather than twice with a worse message.

**Still open from the review above** — I fixed only the security finding:
- `.github/workflows/directory-submission-guard.yml:57` two-dot diff producing false "files outside your submission" accusations once main moves ahead
- `content.config.ts` + `for-ukrainian.plan.md` referencing a `pnpm check:projects` that does not exist
- `astro.config.mjs:13` sitemap exclusion being an unasserted exact string match

Main has also been merged into this branch, so it is no longer behind.

rejifald:
Superseded by #528 — closing this one rather than pushing the redesign on top of it.

#528 carries all five commits from this branch, rebased onto current `main` (this branch had fallen behind and no longer fast-forwards), and then rebuilds the directory page on top of them. Nothing here is dropped: every file this PR touches is present there, including the twenty entries, the `projects` collection schema, the submission guard workflow and CODEOWNERS path, the sitemap exclusion, and the plan document.

The reason for the rebuild rather than a follow-up: the layout this PR shipped was a single 768px column, 5,369px tall, with no index, no counts and no way to jump — so a reader who came for a typeface had to scroll past games and courses to learn whether one was even listed, and the submission invitation this plan calls a day-one feature sat below all of it. #528 replaces it with a sphere-grouped, filterable board at 3,981px, keeping the grouping, the per-locale alphabetical order, and the no-grading rule intact — the last of these now enforced by the interface as well as the data, since every card is the same size and carries the same fields.

Review #528 instead. The branch behind this PR is left in place as the record of how the directory was first built.

PR #459, opened 2026-08-14
fix(audit): cite two vantages that differ in core/serving-decided-by-ip

### LINKED ISSUE #432: audit: core/serving-decided-by-ip cites the first two vantage readings, not two that differ
`core/serving-decided-by-ip` checks that *some* two vantage readings differ, then builds its finding from the *first two* readings regardless. With three or more vantages it can publish a `fail` whose own summary quotes two identical languages.

## What happens

`packages/audit/src/rules/serving.ts:648` filters on a global difference:

```ts
.filter((readings) => distinct(readings.map((r) => r.language)).length >= DIFFERENTIAL_MINIMUM)
```

but `packages/audit/src/rules/serving.ts:613` then takes positional readings:

```ts
const [first, second] = readings;
```

Nothing constrains `first.language !== second.language`.

## Reproduction

Three vantages at one URL under an identical (absent) `Accept-Language` — `kyiv=uk`, `berlin=uk`, `warsaw=en`:

```
verdict: fail
At https://x.ua/ under an identical no Accept-Language, vantage "kyiv" (claims country UA,
unverified) received <html lang="uk"> and vantage "berlin" (claims country DE, unverified)
received <html lang="uk"> — the served language follows the network location, not the
stated header.
```

The verdict is `fail` — a broken promise that breaks a build — and the cited evidence shows no difference at all.

## Why it matters

The rule's own doc calls a vantage's country a claim and words findings carefully so an unverified claim reads as a claim. That care is undone by citing a pair that does not demonstrate the claim being made.

Currently latent: the Node collector is single-vantage, so `multi-vantage` is never derived. It becomes live the moment #421 lands, and `Evidence` is the public API — a merged or third-party bundle can carry three vantages today.

## Fix sketch

Pick the citation pair from the readings, not by position: find the first pair whose `language` differs, and cite those two (keeping the full reading set in `evidence`).

rejifald:
Closes #432

## The defect

`core/serving-decided-by-ip` admitted a vantage leg because *some* two of its readings disagreed:

```ts
.filter((readings) => distinct(readings.map((r) => r.language)).length >= DIFFERENTIAL_MINIMUM)
```

and then built the finding from whichever two readings happened to come first:

```ts
const [first, second] = readings;
```

Nothing tied those two together. With three or more vantages the differing pair sits anywhere in the leg, so the rule could publish a `fail` — a broken promise, the thing that breaks a build — whose own summary quoted one language twice and demonstrated nothing.

This matters more than a cosmetic mis-quote: the rule's doc treats a vantage's country as a *claim* and words findings so an unverified claim reads as a claim. Citing a pair that does not demonstrate the accusation undoes exactly that care.

## Before / after

Three vantages at one URL under an identical (absent) `Accept-Language` — `kyiv=uk`, `berlin=uk`, `warsaw=en`.

**Before** (`verdict: fail`):

> At https://example.com.ua/uk/ under an identical no Accept-Language, vantage "kyiv-proxy" (egress country verified as UA) received `<html lang="uk">` and vantage "berlin-proxy" (claims country DE, unverified) received `<html lang="uk">` — the served language follows the network location, not the stated header.

**After** (`verdict: fail`):

> At https://example.com.ua/uk/ under an identical no Accept-Language, vantage "kyiv-proxy" (egress country verified as UA) received `<html lang="uk">` and vantage "warsaw-proxy" (claims country PL, unverified) received `<html lang="en">` — the served language follows the network location, not the stated header.

## The fix

A new `differingPair(readings)` returns the first pair whose `language` differs, and `decidedByIpFinding` cites that pair. The separate `distinct(...) >= DIFFERENTIAL_MINIMUM` admission filter is gone: `differingPair` returning `null` is what makes a leg yield no draft, so **one call is both the admission test and the citation** and the two cannot drift apart again.

Deliberately unchanged:

- **All three probes stay in `evidence`** — only the sentence narrows to the pair that carries the claim. A reader can still replay the whole leg.
- **Verdict boundaries.** A leg with no differing pair still yields no draft, so the rule still returns `pass()`, not `not-applicable`; the `not-applicable` branch still keys off "fewer than two readings with a declared language", and the rule is still `not-collected` without `multi-vantage`. `distinct` and `DIFFERENTIAL_MINIMUM` both remain in use elsewhere in the file.
- **The summary's wording**, including the claim-vs-measurement care in `vantageLabel`.

## How the regression test proves it

The defect is latent — the Node collector is single-vantage, so `multi-vantage` is never derived — so the test builds three-vantage `Evidence` directly (`Evidence` is the public API; a merged or third-party bundle can carry three vantages today) and drives it through `evaluate` like the surrounding cases.

Three cases in `src/rules/serving.test.ts`:

- **`cites two vantages that actually differ, not the first two positionally`** — the regression test. `quotedDeclarations()` pulls every `<html lang="…">` the summary quotes, in order, and asserts `['<html lang="uk">', '<html lang="en">']`, plus that the cited vantages are kyiv and warsaw and that berlin is not named.
- **`keeps every reading in the evidence while citing only the differing pair`** — guards against the wrong fix, one that narrows `evidence` to the cited pair. Passes both before and after, by design.
- **`does not cite a pair whose tags differ only below the language subtag`** — `uk` vs `uk-UA` is one language, so that pair demonstrates nothing; the comparison is on `language`, not on the raw tag.

Verified as a real regression test by reverting only `src/rules/serving.ts` to `main` and running the suite. It fails with the issue's exact signature:

```
FAIL  src/rules/serving.test.ts > core/serving-decided-by-ip > cites two vantages that actually differ, not the first two positionally
AssertionError: expected [ '<html lang="uk">', …(1) ] to deeply equal [ '<html lang="uk">', …(1) ]

- Expected
+ Received

  [
    "<html lang=\"uk\">",
-   "<html lang=\"en\">",
+   "<html lang=\"uk\">",
  ]
```

and the subtag case fails alongside it, receiving `<html lang="uk-UA">` where it expects `<html lang="en">`.

## Checks

- `pnpm --filter @movar/audit test` — 669 passed (23 files).
- `pnpm validate` — exit 0, including `audit:site`. The dogfood gate is unchanged: `apps/marketing/dist` is filesystem evidence, so this rule reads `not-collected` there as before. No suppression added, none went stale.
- `packages/audit/AGENTS.md` gains a gotcha for the class of bug: a differential rule admitted on an aggregate test must never cite by position.


rejifald:
## Review: BLOCKED on a red gate — the fix itself verifies clean

I checked every claim in the description against the branch rather than taking it on trust, including re-running the pre-fix revert myself. The code change and the regression tests hold up. `metrics-gate` does not.

### Blocking: `metrics-gate` is red, and this PR caused it

The repo's own ruleset (`.github/rulesets/main-metrics-gate.json`) declares `metrics-gate` a required status check, and the failure is the *freshness* check — which `metrics-gate.yml` says explicitly is **never** overridable by `accept-metrics-regression`:

```
✗ Committed coverage is stale or inaccurate.
    committed:  96.4% lines / 90.1% branches
    recomputed: 96.4% lines / 90.2% branches
```

I reproduced it locally and confirmed the drift is this PR's, not something inherited from `main`:

| commit | committed snapshot | recomputed |
| --- | --- | --- |
| `3898c74` — this PR's base | 96.4 / 90.1 | 96.4 / **90.1** — fresh |
| `98d9406` — this PR's head | 96.4 / 90.1 | 96.4 / **90.2** — stale |

`differingPair` plus its three tests moved repo-wide branch coverage by +0.1pp, and `scripts/readme-metrics.snapshot.json` was not refreshed. Worth naming why this was easy to miss: **`pnpm validate` cannot catch it** — it has no `test:coverage` step, so "`pnpm validate` — exit 0" and a red gate are simultaneously true. Every other PR in the current audit batch (#450–#453) passed this gate; only this one moved the rounded number.

To clear it, mirror what CI does and commit the result:

```sh
pnpm test:coverage && pnpm gen:readme --refresh
git add scripts/readme-metrics.snapshot.json   # and README.md, if it moved
```

Locally only `coverage.branches` changes, 90.1 → 90.2; the README metrics block came back "already current". The branch is also one commit behind `main` (now `1f77e57`), so recompute *after* syncing — the snapshot is repo-wide, not package-scoped.

### Everything else verified and passing

**1. The regression tests are real, and the 2-of-3 split is exactly as described.** Reverting only `src/rules/serving.ts` to `main` and leaving the test file intact: **2 failed | 47 passed**, with the issue's signature verbatim —

```
- "<html lang=\"en\">",
+ "<html lang=\"uk\">",
```

`cites two vantages that actually differ…` and `does not cite a pair whose tags differ only below the language subtag` fail; `keeps every reading in the evidence while citing only the differing pair` passes both ways, as the description says. I mutation-tested that one rather than take "guard rail" on faith: narrowing `evidence` to `[first, second]` makes it — and only it — fail. It is load-bearing, not padding.

**2. Root cause, not symptom.** The old admission test `distinct(languages).length >= DIFFERENTIAL_MINIMUM` and the new `differingPair(readings) !== null` are logically the same predicate (both mean "∃ *i*<*j* with differing `language`"), so no leg is newly admitted and none newly dropped — the two just can no longer disagree, because there is now only one of them. Verdict boundaries confirmed unmoved: `not-collected` comes from capability gating (untouched); `not-applicable` still keys off `readings.length >= DIFFERENTIAL_MINIMUM` (untouched); `pass` still means zero drafts. Relocating `first` off position 0 is safe for `subject.url` and `headerLabel` because `vantageLegKey` is `url + header + cookieState`, so both are leg-invariant by construction. All pre-existing boundary tests still pass.

**3. No scope creep.** One commit, three files (`serving.ts`, `serving.test.ts`, `AGENTS.md`), no stray repro script. The test file change is purely additive (+63/−0) — no prettier reformat happened, and none is owed: `prettier --check` is clean on all three.

**4. No new invariant violation.** No `node:` import, no DOM, no clock, no randomness — the addition is a pure loop. Grounding stays `observed` / `fail`, so the `classified`-can-never-fail law is untouched, and `not-collected is never pass` stays pinned by its existing test. On the wording: the summary template and `vantageLabel` are **byte-identical** to `main`, so a vantage's country still reads as a claim, and the new test asserts `claims country PL, unverified` verbatim alongside `egress country verified as UA`.

**5.** `audit-site`, `verify`, `build`, `e2e-offline`, `ci-gate` all green. `apps/marketing/audit-suppressions.json` untouched, nothing added, nothing stale — as claimed. Local `pnpm --filter @movar/audit test` reproduces 669 passed / 23 files, and typecheck + lint are clean.

### Non-blocking nit, worth folding in while you're here

There is no test pinning the three-vantage *negative*: `fromThreeVantages('uk', 'uk-UA', 'uk')` should still be `pass`. That is the boundary where `differingPair` must return `null` across three readings — the exact case the old aggregate filter also got right, so a wrong rewrite could silently start failing it. The two-vantage version is covered; the n=3 one is not.

Re-request review once the snapshot is refreshed and `metrics-gate` is green.

Issue #431, opened 2026-08-14
audit: a non-numeric --budget becomes NaN and silently disables the hard request ceiling


rejifald:
`parseArgs` in `packages/audit/src/collect/cli.ts` converts `--budget` with `Number()` and never validates the result. `--budget lots` yields `NaN`, which silently removes the hard request ceiling the ADR requires.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

`packages/audit/src/collect/cli.ts:78`:

```ts
...(budget === undefined ? {} : { budget: Number(budget) }),
```

`NaN` then defeats both guards:

- `packages/audit/src/collect/probe.ts` — `if (spent >= budget)` is false for every `spent`, so `RequestBudgetExhaustedError` never throws.
- `packages/audit/src/collect/node.ts` — `remaining()` returns `Math.max(0, NaN - spent)` = `NaN`, and the collector's `if (prober.remaining() === 0) break` never fires.

## Reproduction

```
spent: 500 | remaining(): NaN | remaining()===0: false
```

500 sequential requests against a live third-party site, from a typo in a flag value.

## Why it matters

`probe.ts`'s module doc: "**A hard, sequential request budget**, enforced here rather than trusted to callers." It is enforced here — unless the caller hands it `NaN`, in which case it silently isn't. This is an abuse-posture defect against sites the tool is pointed at without their involvement.

## Also in `parseArgs`

`value()` returns the next argv element unconditionally, so `movar-audit --url --follow` parses as `url === '--follow'` and then fetches it.

## Fix sketch

Validate in `parseArgs`: reject a `--budget` that is not a non-negative integer, and reject a flag value that itself starts with `--`. Both are exit-code `2` conditions ("the run never happened"), which the CLI already models.

Issue #434, opened 2026-08-14
audit: robots.txt wildcards are matched literally and grouped user-agent lines are dropped


rejifald:
`robotsAllows` matches `Disallow` patterns with `path.startsWith(pattern)`, so RFC 9309's `*` and `$` metacharacters are treated as literal characters and the rules they express never apply. Separately, `parseRobots` mishandles grouped `User-agent` lines.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## A — wildcards are literal

`packages/audit/src/collect/probe.ts:541`:

```ts
patterns.filter((pattern) => path.startsWith(pattern))
```

RFC 9309 §2.2.3 defines `*` (any sequence) and `$` (end of match) as special in a path pattern. Reproduction:

```
parsed: {"disallow":["/*/private","/*.json$"],"allow":[]}
/uk/private allowed? true
/data.json  allowed? true
```

Both should be disallowed. `Disallow: /*?` — very common, to block query strings — is likewise inert.

## B — grouped user-agent lines depend on order

`packages/audit/src/collect/probe.ts:530` overwrites the flag per line:

```ts
inWildcardGroup = directive.value === '*';
```

RFC 9309 §2.2.1: consecutive `User-agent` lines form **one** group applying to all of them. So this parses correctly:

```
User-agent: Googlebot
User-agent: *
Disallow: /private
```

and this drops the group entirely:

```
User-agent: *
User-agent: Googlebot
Disallow: /private
```
```
parsed: {"disallow":[],"allow":[]}
```

## Why it matters

Both defects fail open — the collector fetches paths the site asked it not to. `parseRobots`' doc says "honouring the wildcard is the conservative reading"; these two gaps mean it is honoured only for patterns with no metacharacters, written in one of two orders.

## Fix sketch

Compile each pattern to an anchored matcher (`*` → `.*`, trailing `$` → end anchor, everything else escaped), keeping longest-match-wins on the **pattern length** as now. For B, accumulate consecutive `User-agent` values and open the group if any of them is `*`.

Issue #433, opened 2026-08-14
audit: a redirect chain longer than maxHops is recorded as an error, so core/switch-bounces never sees it


rejifald:
When a redirect chain exceeds `maxHops`, `walk` returns `response: null`, `resolveOutcome` maps that to `outcome: 'error'`, and `adjudicableProbes` drops the probe entirely. The recorded `redirectChain` is right there in the evidence and no rule ever sees it.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

`packages/audit/src/collect/probe.ts:369` — falling out of the hop loop returns a `Hop` with `response: null` and `transportError: false`. `resolveOutcome` then keys only on `response === null`:

```ts
if (transportError || response === null) return 'error';
```

`packages/audit/src/capability.ts:adjudicableProbes` keeps only `outcome === 'ok'`.

## Reproduction

An 11-hop chain of distinct URLs (no cycle, so the `seen` set never fires), `maxHops: 10`:

```
outcome: error | status: 0 | hops recorded: 11
probes a rule can see: 0
capabilities derived: []
```

11 requests spent; zero adjudicable evidence; `http` not even derived.

## The inconsistency

A **cyclic** chain exits early through the `seen` check at `packages/audit/src/collect/probe.ts:394`, returns a live response, and stays `'ok'` — fully adjudicable. So a 2-hop loop is evidence and an 11-hop chain is discarded, which is backwards.

## Why it matters

`probe.ts`'s module doc: "**Redirect chains are walked, not delegated.** `redirect: 'manual'` and one hop at a time, because `core/switch-bounces` — the rule this product exists for — is adjudicated entirely from the recorded chain." On the most pathological chain there is, that rule reports `not-collected`.

## Fix sketch

Distinguish "too many hops" from "no response". Either add a `ProbeOutcome` for it, or return the last redirect response (as the `unresolvable` branch already does) so the chain stays adjudicable and `core/switch-bounces` can read it.

Issue #429, opened 2026-08-14
audit: the collector applies the first origin's robots.txt to cross-origin declared targets, and never reads theirs


rejifald:
`resolveRobots` in `packages/audit/src/collect/node.ts` fetches `robots.txt` from the origin of the URL the operator typed, and `followDeclared` then applies those rules to **every** declared target — including hreflang alternates on another origin, which is the standard cross-domain locale pattern (`brand.de` / `brand.pl` / `brand.ua`).

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

`packages/audit/src/collect/node.ts:307`:

```ts
if (!robots.allows(new URL(target).pathname)) continue;
```

`robots` was built from `options.url`'s origin. `target` may be any origin. Both directions are wrong, and both reproduce:

**A — the wrong site's rules block a legitimate target, which then reads as a defect.**
With `example.com/robots.txt` = `User-agent: *\nDisallow: /de`, a declared alternate `https://example.de/de/` is skipped. `core/hreflang-target-unresolvable` then publishes a `fail`:

> The hreflang="de" target "https://example.de/de/" is absent from the collected page set, so the declared alternate cannot be reached.

example.de serves that page perfectly well and its own `robots.txt` permits it.

**B — the other origin's robots.txt is never read at all.**
With a permissive `example.com/robots.txt`, the collector fetched `https://example.de/de/` and the request log shows `https://example.de/robots.txt` was never requested:

```
fetched: [ 'https://example.com/', 'https://example.com/robots.txt',
           'https://example.de/de/', 'https://example.com/' ]
```

## Why it matters

Direction A is a **false accusation about a named company** — `docs/movar-audit.md` identifies that as the one failure mode that ends the product.

Direction B breaks the module's own stated posture. `resolveRobots`' doc comment says `robots.txt` is "honoured for declared-target expansion, which is automated multi-page access". For every origin except the first, it is not honoured, because it is never fetched.

## Fix sketch

Resolve and cache `robots.txt` per origin, keyed on `new URL(target).origin`, and evaluate each target against its own origin's rules. The per-origin fetch costs a request against the budget, which is the honest price of the posture the ADR describes.

PR #461, opened 2026-08-14
fix(audit): validate the prober's ceilings at construction, not at first use

### LINKED ISSUE #456: audit: createProber accepts a NaN budget, so "enforced here, not trusted to callers" holds only for the CLI path
`probe.ts`'s module doc says the request budget is "enforced here rather than trusted to callers", but `createProber` accepts any `number` and applies no validation of its own. A `NaN` budget disables the ceiling completely.

Surfaced while reviewing #450; verified against `3898c74`.

## Reproduction

```
createProber({ vantage, budget: NaN, fetchImpl })  then 60 probes
  -> spent=60  remaining=NaN  remainingIsZero=false
```

Both guards fail open: `spent >= NaN` is never true so `RequestBudgetExhaustedError` never throws, and `remaining()` returns `NaN`, which never satisfies the collector's `if (prober.remaining() === 0) break`.

## Relationship to #431

#431 fixed this at the **CLI edge** — `parseArgs` now rejects a `--budget` that is not a non-negative safe integer, so the shipped path is safe today. This issue is about the claim `probe.ts` makes for itself. From its module doc:

> **A hard, sequential request budget**, enforced here rather than trusted to callers. Exhaustion throws `RequestBudgetExhaustedError`; it is never a quietly-shortened run, because a tool that silently bounds coverage reads as "covered everything".

That is an argument for a guard in `createProber`, and there isn't one. Any second caller — the Safari host app collector (#422), a future relay, a test harness — gets no protection.

Latent today: `createProber` is not exported from `src/index.ts`, only via the `./*` subpath, and the only in-repo caller is `collectNetwork`.

## Fix sketch

Validate in `createProber`: reject a `budget` that is not a non-negative safe integer, throwing at construction. Same for `maxHops` and `timeoutMs`, which have the same shape. Constructing with an unusable budget is a programming error, so throwing is right — unlike the CLI edge, where it is user input and belongs in an exit code.

rejifald:
`probe.ts` opens by claiming the request budget is "enforced here rather than trusted to callers". `createProber` accepted any `number` and applied no validation of its own, so the claim held only for the CLI path.

## The defect, re-verified on `3898c74`

Branched from fresh `origin/main` and reproduced before touching anything:

```
NaN budget:      constructed OK; spent=60 remaining=NaN remainingIsZero=false
Infinity budget: constructed OK; spent=60 remaining=Infinity remainingIsZero=false
-5 budget:       threw RequestBudgetExhaustedError after 0 probes
1.5 budget:      threw RequestBudgetExhaustedError after 2 probes
NaN maxHops:     constructed OK; spent=0  (walk never fetched at all)
NaN timeoutMs:   Node TimeoutNaNWarning — "Timeout duration was set to 1"
```

Both budget guards fail open on `NaN`: `spent >= NaN` is false for every `spent`, so `RequestBudgetExhaustedError` never throws, and `remaining()` is `NaN`, which never satisfies the collector's `if (prober.remaining() === 0) break`. `Infinity` uncaps a run just as completely; a negative budget exhausts before the first request, and a fractional one truncates at an arbitrary point.

`maxHops` and `timeoutMs` have the same shape and were checked too, since the issue asked for them. Both fail in their own way, and both fail silently:

- **`maxHops: NaN`** makes `hop <= maxHops` false on the first pass, so `walk` returns without fetching at all. The probe then reports `outcome: 'error'`, `status: 0` for a request nobody ever made — a fabricated transport failure.
- **`timeoutMs: NaN`** is coerced to `1` by `setTimeout`, so every request aborts after 1 ms. Node warns about it; nothing else does.

## Scope — still latent, including after #422

#431 fixed the same class of bug at the **CLI edge** (`parseArgs` rejects a `--budget` that is not a non-negative safe integer), so the shipped path was already safe. This is about the claim the module makes for itself.

One correction worth recording, since #422 landed during review and was expected to make this concrete: **it did not.** The Safari host app collector took the native path — `COLLECTOR_ID = 'swift-urlsession'`, with its own budget and hop cap enforced in `AuditProbe.swift` — and imports only `@movar/audit/collect/assemble` and `collect/digest-dom`. It never constructs a `Prober`. `collectNetwork` remains the only in-repo `createProber` caller, so the guard is still latent exactly as #456 said. It protects the next direct caller: a relay, a test harness, a future in-process collector.

## The fix

`createProber` now resolves each of `budget`, `maxHops`, and `timeoutMs` through one guard: a non-negative safe integer, or a `TypeError` at construction.

Thrown rather than defaulted or clamped, and thrown at construction rather than at the first probe — a prober that cannot enforce its own ceiling must not exist, rather than exist and fail open on first use. Unlike the CLI edge, where `--budget` is user input and belongs in an exit code, a caller in this position is code, and a silently corrected ceiling is the same quietly-bounded run the ADR's "no silent caps" rule exists to prevent.

Zero stays legal for all three — spend nothing, follow no redirect, abort at once — so the guard is a range check and never a falsy one. `parseArgs` already accepts `--budget 0`, and rejecting it here would have broken that path.

## Error type: a plain `TypeError`

Considered, and rejected, in order:

- **Reusing `RequestBudgetExhaustedError`** would conflate two different conditions. That error means a run *reached* its ceiling — a real, expected outcome a caller may legitimately catch to report partial coverage. Routing a construction bug through the same class means that handler swallows it, which is exactly the failure this PR is fixing, one level up. A test pins that the two stay distinguishable.
- **A new exported error class** would put a name on the package's public surface that can never be renamed, and would *invite* the `catch` that must not happen. Recovering from a malformed ceiling by degrading is not a recovery. Nothing should branch on this, so it earns no exported name.
- **`TypeError`** is the platform's own name for "you handed this function a value it cannot use", and it covers both rejected shapes under one guard — `NaN` / `Infinity`, which are not usable numbers, and `-1` / `1.5`, which are numbers outside the allowed set. A single predicate with a single error reads better here than a `TypeError` / `RangeError` split.

## Tests

Seven new tests in `src/collect/probe.test.ts`, under `describe('createProber option validation')`. Six failed before the fix and pass after; the seventh is the over-strictness guard and passes both ways by design (it would have caught a falsy check or a rejected default).

| Test | Before |
| --- | --- |
| `rejects a NaN budget at construction, since NaN silently uncaps the run` | fail |
| `rejects every budget that is not a non-negative safe integer` | fail |
| `rejects a malformed maxHops or timeoutMs on the same terms` | fail |
| `refuses a malformed budget with a TypeError, never RequestBudgetExhaustedError` | fail |
| `names the offending option and value, since construction is refused over it` | fail |
| `guards a caller the type system does not` | fail |
| `accepts zero for each option, and the defaults when they are omitted` | pass (by design) |

## Verification

| Command | Result |
| --- | --- |
| `pnpm --filter @movar/audit test` | 25 files, 736 tests passed |
| `pnpm nx run marketing:audit --skip-nx-cache` | exit 0 — findings and suppressions unchanged |
| `pnpm check:readme` | clean |
| CI | `ci-gate`, `metrics-gate`, `verify`, `build`, `audit-site`, `e2e-offline` all pass |

## Merges and the metrics snapshot

The branch carries merges of `main` through #475, so `cli.ts` and `cli.test.ts` are byte-identical to `main` — this PR still changes neither.

The snapshot commits are bookkeeping, kept separate from the fix. Two things are worth a reviewer's attention:

**`loc` 60518 → 60347.** This is a correction, not growth. `BUILD_DIRS` in `scripts/gen-readme-metrics.mts` does not exclude `.astro`, so any `pnpm metrics` run on a tree where `astro sync` had run counted `apps/marketing/.astro/content.d.ts` (208 lines) and `types.d.ts` (2) as source — which is why `main` ships an inflated figure. Regenerated here with that directory absent: `main`'s clean 60308 + 39 from this branch's `probe.ts` = 60347. The README badge stays `60.3k`. Filed as #472; avoided here, not fixed. An earlier revision of this PR committed a `loc` measured on an inflated tree and described the increase as unrelated drift — that was wrong, and is retracted.

**`lines` 96.6, `branches` 90.3 — both unchanged by the #475 update.** Measured rather than assumed, and now three times. On this machine `main` at `0f2d6c7` recomputed to 6069/6286 = 96.5479 %, matching the 96.5 it already carried and its own gate had validated, so the base agrees; this branch then measured 6073/6290 = 96.5501 %, that base plus four always-covered lines from `requireCount` and `shown`. After merging #475 it is 6076/6293 = 96.5517 % lines and 4058/4495 = 90.2781 % branches — the same rounded pair, independently reproduced. The line figure clears the 96.55 % boundary by 0.0017 points, which is thin enough to keep flagging; it is a property of one-decimal storage rather than of this change, and `metrics-gate` recomputes from scratch each run, so a flip would fail loudly rather than ship a wrong badge.

Closes #456


rejifald (commented):
Reviewed adversarially against `3898c74` and re-verified independently rather than from the description. **Five of seven checks pass cleanly. The fix itself is correct, well-scoped, and well-argued — I have no objection to the code.** Two things block: the branch now conflicts with `main`, and the `loc` field in the snapshot is wrong (and its attribution in the description is not accurate).

---

## ✅ 1. Regression tests fail without the fix

Reverted **only** `src/collect/probe.ts` to `3898c74`, kept the tests. The claimed 6/1 split reproduces exactly:

```
× rejects a NaN budget at construction, since NaN silently uncaps the run
× rejects every budget that is not a non-negative safe integer
× rejects a malformed maxHops or timeoutMs on the same terms
× refuses a malformed budget with a TypeError, never RequestBudgetExhaustedError
× names the offending option and value, since construction is refused over it
× guards a caller the type system does not
✓ accepts zero for each option, and the defaults when they are omitted
```

The over-strictness guard is genuinely load-bearing — I mutated the guard two ways and it is the **only** test that catches either:

* `if (!value || !Number.isSafeInteger(value) || value < 0)` — falsy check → the other six go green, that one fails alone.
* validating before the `??` default (so an omitted option is refused) → it fails again, alongside the whole pre-existing `createProber` suite.

And the `--budget 0` path is real: `DIGITS_ONLY = /^\d+$/` matches `"0"`, `Number.isSafeInteger(0)` is true, so `parseArgs` accepts it and `collect()` forwards it to `collectNetwork` → `createProber({ budget: 0 })`. A falsy check would have broken the shipped CLI. Good test.

## ✅ 2. The fix closes the hole

Exercised all three options against 8 rejected values each (24 combinations) — `NaN`, `±Infinity`, `-1`, `-0.5`, `1.5`, `MAX_SAFE_INTEGER + 1`, `-5` — every one throws `TypeError` at construction for `budget`, `maxHops` and `timeoutMs`. `0` and safe positives are accepted, and `budget: 0` still exhausts on the first probe with zero fetches. Public surface is unchanged (`git diff` of `^export` lines base→head is empty), so "it earns no exported name" holds literally.

Two non-blocking observations:

* `null` for any of the three is still accepted and silently becomes the default — `??` is nullish-coalescing, so `null` reads as "omitted". Not the defect class (you get a real, enforced ceiling of 40, not an uncapped run), but for a guard whose stated purpose is "callers the type system does not cover", `null` is about the second most likely untyped mistake after a string.
* `shown([])` renders as the empty string, so the message ends `…not ` with nothing after it. Cosmetic.

## ✅ 3. Both extra defects were real, and are covered

Reproduced both against the pre-fix module before trusting the writeup:

```
PREFIX budget NaN      -> fetches=60 spent=60 remaining=NaN remainingIsZero=false
PREFIX budget Infinity -> spent=60 remaining=Infinity
PREFIX budget -5       -> RequestBudgetExhaustedError after spent=0
PREFIX budget 1.5      -> threw after spent=2
PREFIX maxHops NaN     -> fetches=0 spent=0 outcome=error status=0 chain=[]
PREFIX timeoutMs NaN   -> TimeoutNaNWarning: "Timeout duration was set to 1." outcome=error
PREFIX timeoutMs 15000 -> outcome=ok      (control, same slow fetch)
```

`maxHops: NaN` really does fabricate a transport failure for a request nobody made — zero fetches, `outcome: 'error'`, `status: 0`, empty chain. `timeoutMs: NaN` really is coerced to 1 ms; the control run proves the same fetch succeeds at 15 s. Both are in scope (#456 asked for these two on the same terms) and neither was used to widen the change: the `probe.ts` diff is exactly three hunks — module doc, `shown` + `requireCount`, and the three call sites.

## ✅ 4. The error type

The reasoning holds up. `RequestBudgetExhaustedError extends Error`, so a `TypeError` is not an instance of it and the two stay distinguishable; and the argument is not hypothetical — `collectNetwork`'s robots path has a bare `catch {}` that returns `{ posture: 'honoured', allows: () => true }`, exactly the shape of handler that would swallow a construction bug routed through the budget class. `createProber` is called outside any `try` in `collectNetwork`, and `runCli` does not wrap `collect()`, so the `TypeError` surfaces loudly. Correct.

The pin is load-bearing: mutating the guard to `throw new RequestBudgetExhaustedError(value)` fails six tests including `refuses a malformed budget with a TypeError, never RequestBudgetExhaustedError`.

Minor, not a change request: `RangeError` is arguably the platform's idiom for a numeric value outside an allowed range (`new Array(-1)`, `(1).toFixed(101)`), which would cover `-1` / `1.5` better than `TypeError`. You considered and rejected the split with a defensible rationale, nothing branches on the class, and `TypeError` is right for the headline `NaN` case — so I'd leave it.

## ⚠️ 5. The metrics snapshot

**Coverage: the numbers are right, and the gate was *not* just told what it wanted.** Worth being clear about this, because the description is harder on itself than the facts warrant. `metrics-gate` recomputes coverage from a fresh `pnpm test:coverage` in CI and compares it against the pre-refresh copy of what the PR committed; a value that did not match would have exited 2. The run on `1a38801` printed `✓ No metrics regression. Coverage: 96.4% lines / 90.2% branches.` — an independent recomputation that agreed. The number was verified, just not by you.

Your darwin figure reproduces exactly on my machine:

```
AGGREGATE lines    5645/5856 = 96.3969%  -> 96.4
AGGREGATE branches 3711/4117 = 90.1385%  -> 90.1     (3712 would give 90.163 -> 90.2)
```

One covered branch short of the boundary, precisely as described. Likely cause, though: `.nvmrc` pins Node 22 and the runner used **v22.23.2**, while a current darwin dev box is on Node 24 (mine is v24.18.1) — V8's coverage counts differ across versions, so "could not be produced locally" is really "not on an unpinned Node", which is a smaller claim.

And it stops mattering once the branch is updated. I merged current `main` locally and re-ran the full suite:

```
AGGREGATE lines    5658/5869 = 96.4048%  -> 96.4
AGGREGATE branches 3723/4127 = 90.2108%  -> 90.2
```

Both committed values become reproducible on darwin, two branches clear of the boundary. (`main` had already moved branches to 90.2 on its own via #462, and has since moved again — see below.)

**`loc`: this is the part I'd like changed.** The committed `56922` *is* what the tooling emits — but only on a tree where Astro has been synced. I reproduced it exactly:

```
clean checkout at 1a38801        -> loc = 56712
after `astro sync`               -> loc = 56922   (+210)
  208  apps/marketing/.astro/content.d.ts
    2  apps/marketing/.astro/types.d.ts
```

`BUILD_DIRS` in `scripts/gen-readme-metrics.mts` lists `node_modules .output dist .nx .wxt coverage .turbo` — **not `.astro`** — so `scanSourceStats` counts two generated, gitignored `.d.ts` files as source. That is a pre-existing tool bug (older snapshots carry the same inflation: +206 at #406, +469 at #424, +176 at #427), not something this PR introduced. But this is the first time it crosses a rendered boundary: the README goes `56.7k` → `56.9k`, and the true source count at this head is 56712, which still renders `56.7k`. So the badge, whose own text says "TypeScript under `apps/` + `packages/`, tests excluded", becomes wrong.

The attribution in the description — *"that field has no freshness check, so it had drifted across the PRs landing since #427; this is just the current count"* — is also not accurate. Breaking down the committed +205:

| | lines | what it is |
| --- | --- | --- |
| drift from #449 / #425 / #450 / #451 / #452 | +132 | genuinely not your work |
| **this PR's own `probe.ts`** | **+39** | your work (`3898c74` 56673 → `1a38801` 56712) |
| growth of the generated `.astro/content.d.ts` | +34 | not source at all |

It is *not* scope creep — `gen:readme --refresh` rewrites the whole snapshot, so you cannot refresh coverage without also refreshing `loc`, and carrying it is unavoidable. But it isn't "not your own work" either, and the resulting badge is wrong. Since the snapshot has to be regenerated anyway (below), please regenerate it on a tree without `apps/marketing/.astro/` present, or after removing it.

## ✅ 6. Scope and invariants

Five files, all justified: `README.md` (1 line), `packages/audit/AGENTS.md` (+1/-0, pure append, no reflow), `probe.test.ts` (+110/-0, pure addition — no existing test touched), `probe.ts` (3 hunks), `readme-metrics.snapshot.json`.

* `src/collect/cli.ts` — untouched. Confirmed, as are `suppress.ts`, `evaluate.ts`, `report.ts`, `locator.ts`, `serving.ts`, `rules/ua.ts` (the in-flight set).
* **#429 / #433 / #434 verified behaviourally, not textually.** I ran a differential harness with base `probe.ts` and head `probe.ts` side by side: `parseRobots` + `robotsAllows` over a 10-document / 13-path corpus (wildcards, grouped `User-agent` lines, comments, longest-match ties), `resolveLocation`, `isChallengeResponse`, `sha256`, the exported constants, request headers, and eight redirect-chain scenarios — plain 200, three-hop chain, redirect loop, unresolvable `Location`, challenge-at-200, `maxHops: 0`, and two over-`maxHops` endless chains. Byte-identical probe records and identical fetch sequences in every case. In particular #433's fall-through (`outcome=error status=0` when the hop cap is exhausted) is preserved verbatim — the fix stops `maxHops: NaN` from *reaching* that path without altering the path itself. That is the right line to draw.
* Kernel purity intact: no new imports at all, `src/purity.test.ts` still excludes `src/collect/`, and `node:crypto` is not on its forbidden list. `pnpm check:readme`, `--filter @movar/audit typecheck` and `lint` all clean.

## ❌ 7. CI green on a current head

This is the blocker.

```
$ gh pr update-branch 461
X Cannot update PR branch due to conflicts
```

GitHub now reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. The branch is based on `3898c74`; since then #453, #462, #459 and #422 have all landed. The conflict is `scripts/readme-metrics.snapshot.json` — both sides rewrote `loc` — and `README.md` and `packages/audit/AGENTS.md` have also moved on `main` (the AGENTS.md append lands at the same tail). `src/collect/probe.ts` itself is unchanged on `main`, so the fix will still apply cleanly.

The green checks on this PR ran against `3898c74`, and under strict mode that is not evidence for the merged result. It also means the snapshot is now doubly stale: `main` currently carries `loc 60398` and coverage `96.5 / 90.3` (#459 and #422 landed after my measurements above), so the committed `96.4 / 90.2` will fail the freshness check on any updated head.

Worth noting that #422 — the Safari host app collector this PR names as the second caller that "gets no protection" — has now landed, so the latent case is one step closer to real.

---

### To unblock

1. Merge or rebase onto current `main` and resolve the three conflicts.
2. Re-run `pnpm metrics` on the new head, **without `apps/marketing/.astro/` present**, and commit the refreshed snapshot + README.
3. Optionally, fix the underlying `loc` bug separately by adding `.astro` to `BUILD_DIRS` in `scripts/gen-readme-metrics.mts` — happy to see that as its own PR rather than here.

Nothing above asks for a change to the fix. Checks 1–4 and 6 pass on their merits, and the `TypeError` argument in particular is the right call for the right reasons.

rejifald (commented):
Re-reviewed `db2b686`. **Both blockers are genuinely cleared and the substance is approved — I was ready to merge.** `main` moved to `47df24f` (#475) between my verification and the merge, so the PR is BEHIND and `CONFLICTING` again on the shared snapshot. I did not merge a stale base. Everything below is measured, and the numbers you need to land it are at the bottom.

### Blocker 1 — conflicts and `cli.ts`: cleared

`origin/main` was an ancestor of `db2b686` when I checked; diff is exactly the 5 files. `cli.ts` and `cli.test.ts` are identical to main by **blob OID and mode**:

```
main   100644 blob aefcd022…  cli.ts      100644 blob 2af9d4b8…  cli.test.ts
db2b686 100644 blob aefcd022…  cli.ts      100644 blob 2af9d4b8…  cli.test.ts
```

No mode change anywhere in the diff (`git diff --raw` shows `100644 100644` on all five). I reproduced the 644→755 flip on my own box — `pnpm install` chmods it because `packages/audit/package.json` declares it as `bin` — so that was correctly diagnosed as install dirt, not a change, and the commit is clean.

`probe.ts` (`c8527cc`) and `probe.test.ts` (`bc0f6d07`) are **byte-identical to the head I already verified**, and main's `probe.ts` is still `553b556` — the exact blob my behavioural differential ran against. So checks 1–4 and 6 carry over verbatim; nothing needed redoing. 25 files / 744 tests pass, `check:readme` / `typecheck` / `lint` clean.

### Blocker 2 — `loc`: cleared, and the retraction is there

Reproduced `60285` exactly, twice — a git-tree scan and the verbatim `scanSourceStats` over the working tree, with `apps/marketing/.astro/` absent. Main's clean tree is 60246, so 60246 + 39 (this branch's `probe.ts`) = 60285, and 60456 − 210 + 39 = 60285 checks out. `60.3k` is now the right badge, and this PR is the first snapshot in a while that isn't carrying the `.astro` inflation. Retraction is present verbatim in the body, and #472 files the root cause. Good.

### The #422 correction: verified, the brief was wrong

Confirmed independently, and the corrected bullet is true:

* `git grep 'createProber('` over non-test TS returns exactly one call site — `src/collect/node.ts:67`. `collectNetwork` is still the only in-repo caller.
* The host app's TS imports only `@movar/audit`, `/evidence`, `/collect/assemble`, `/collect/digest-dom`, `/artifact` — never `collect/probe`.
* `COLLECTOR_ID = 'swift-urlsession'` at `apps/safari-host-app/src/audit/collect.ts:58`, and `AuditProbe.swift` carries its own `defaultRequestBudget = 40` / `maxRequestBudget = 60` / `maxHops = 10`.

The guard stays latent exactly as #456 said. Thanks for catching that rather than inheriting it.

### Coverage: thin, but correct — not a reason to block

Reproduced on my box (darwin, Node 24) to the digit:

```
lines    6073/6290 = 96.5501%  -> 96.6
branches 4056/4493 = 90.2738%  -> 90.3
```

Matching your run and CI's own recomputation on `db2b686`. Three independent measurements agree, so this is a measured value, not an asserted one — and `metrics-gate`'s freshness check recomputes from scratch and exits 2 on a mismatch, so it cannot be satisfied by being handed a number it likes.

On the margin: it is one covered line, and yes that is thin. I don't think it blocks, for three reasons. It is endemic to storing coverage at one decimal, not something this PR created — `main` sits one line on the *other* side of the same boundary (6069/6286 = 96.5479%). This PR's own effect is +4 always-covered lines, which pushes coverage up, not toward an edge. And if a later landing flips it, the failure is loud and self-correcting: `metrics-gate` exits 2 telling the author to run `pnpm metrics`, rather than shipping a wrong badge. Blocking here wouldn't fix the design, it would just hand the same coin-flip to the next PR.

### e2e: the box, not the change

Satisfied. `e2e-offline` passes on `db2b686` on Linux; the PR touches no file under `apps/`; the failure is in `apps/e2e/src/fixtures/extension.ts`'s `waitForServiceWorker`, the headed-Chromium MV3 fixture. The hook fired only because the merge carried main's `apps/extension` / `apps/e2e` changes into the push and they match its `glob` — it is gated on files arriving in the push, not on files this PR authored. The hook's own comment already treats Linux CI as authoritative for e2e baselines.

### What is left — one line

`main` is now `47df24f` (#475), which touched `AGENTS.md` and the snapshot. `gh pr update-branch 461` → `X Cannot update PR branch due to conflicts`; GitHub reports `CONFLICTING` / `DIRTY`. The only conflict is `loc` in the snapshot again (main 60518 vs 60285). `AGENTS.md` auto-merges cleanly — main's insert is mid-list, yours is a tail append, both survive.

I merged `47df24f` locally and re-measured so you don't have to guess:

```
lines    6076/6293 = 96.5517%  -> 96.6     (unchanged — still clears, still by one line)
branches 4058/4495 = 90.2781%  -> 90.3     (unchanged, 5 branches of margin)
loc      60347                             (clean tree, .astro absent)
```

**So the coverage pair you committed survives the update untouched — only `loc` moves, 60285 → 60347, and the README to `60.3k` (unchanged).** Resolve the snapshot conflict by taking your coverage and setting `loc: 60347`, push, and it should go green in one pass. (Cross-check: main's clean tree is 60308, +39 for this branch's `probe.ts` = 60347; main's committed 60518 is 60308 + 210, i.e. #475 is carrying the `.astro` inflation #472 describes.)

Ping me when it is up to date and I will merge — the review itself is settled, and I don't need to re-verify anything but the SHA and the gate.

Issue #448, opened 2026-08-14
audit: the CLI never prints notApplicableReason or missingCapabilities


rejifald:
`formatResult` prints a rule's id and its findings and nothing else. `notApplicableReason` and `missingCapabilities` are both present in the `Report` and neither reaches stdout — on a typical run that is two thirds of the catalogue going silent with no explanation.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

`packages/audit/src/collect/cli.ts:88`:

```ts
function formatResult(result: RuleResult): string {
  return `   ${result.rule}\n${result.findings.map(formatFinding).join('')}`;
}
```

## Reproduction

A single-page filesystem run — 28 of 41 rules did not run. The whole of what the operator sees about them:

```
coverage: 7 ran · 17 n/a · 11 not-collected
– not-applicable (17)
? not-collected (11)
```

while the report itself holds, for those same rules:

```
example reason  : "the URL carries no language marker"
example missing : ["site"]
```

## Why it matters

`cli.ts`' module doc: *"It prints what a rule decided and why, which is what makes a real run diagnosable."* It prints the what. The why is dropped exactly where it is most needed — an operator seeing 11 `not-collected` rules has no way to learn that adding a second page, or `--follow`, would run them.

## Fix sketch

Append the reason to the rendered line:

```ts
const why =
  result.notApplicableReason ??
  (result.missingCapabilities === undefined ? undefined : `needs ${result.missingCapabilities.join(', ')}`);
return `   ${result.rule}${why === undefined ? '' : ` — ${why}`}\n${…}`;
```

PR #463, opened 2026-08-14
fix(audit): count the pages a rule judged, so 1 of 3 stops reading as a clean pass

### LINKED ISSUE #445: audit: per-page adjudication coverage is invisible — 1 of 3 pages reads as a clean pass
`mergeOutcomes` collapses a page-scoped rule's per-page outcomes so that any single passing page yields `verdict: 'pass'`. A rule that actually adjudicated one of three pages is indistinguishable in the report from one that adjudicated all three.

## What happens

`packages/audit/src/evaluate.ts:78` sets `sawPass` on any passing page; `packages/audit/src/evaluate.ts:113` returns `pass` whenever it is set, discarding how many pages were `not-applicable`.

## Reproduction

A page rule that passes on `page-1` and is `not-applicable` on `page-2` and `page-3`:

```
verdict: pass | coverage: {"rules":1,"ran":1,"notApplicable":0,"notCollected":0,
                           "passed":1,"failed":0,"warned":0}
```

Nothing in the report can express "this rule looked at 1 of your 3 pages".

## Why it matters

`AGENTS.md` and `evaluate.ts` both lead with: *"`not-collected` is never `pass`. Silently passing what it did not check is the default failure mode of every audit tool."*

That invariant holds at rule granularity and is inverted at page granularity, where the roll-up is a pass and `CoverageSummary` has no vocabulary for per-page adjudication at all.

## Also

`reason ??= outcome.reason` keeps only the first page's `notApplicableReason` and discards the rest, so an all-not-applicable rule reports one page's reason as though it were the rule's.

## Fix sketch

Carry per-page counts on `RuleResult` (`pagesAdjudicated` / `pagesNotApplicable`) and surface them in `CoverageSummary`, so a rule that skipped most of the site says so on the report's face.

rejifald:
`mergeOutcomes` reduced a page rule's per-page outcomes to a `sawPass` flag, and `runRule` returned `pass` whenever it was set. A rule that adjudicated one page of three and found nothing to judge on the other two reported `verdict: pass`, and `CoverageSummary` had no vocabulary for per-page adjudication at all:

```
verdict: pass | coverage: {"rules":1,"ran":1,"notApplicable":0,"notCollected":0,
                           "passed":1,"failed":0,"warned":0}
```

The module's own invariant — "`not-collected` is never `pass`; silently passing what it did not check is the default failure mode of every audit tool" — held at rule granularity and was inverted one level down.

**Re-verified against current `main` (`1f77e57`, 46 rules, `Evidence` v2) before touching anything.** It still reproduces, byte for byte with the issue.

It is not hypothetical either. On movar.fyi's own dogfood run:

| rule | verdict | pages judged | pages skipped |
| --- | --- | --- | --- |
| `core/lang-contradicts-url` | `pass` | 11 | 9 |

18 of the page rules skip at least one page; site-wide the run makes 394 adjudications against 186 (rule, page) slots with no subject. The report said none of it.

## What changed

**Count instead of flagging.** `RuleResult` gains `pagesAdjudicated` / `pagesNotApplicable` — *distinct pages*, present iff a page-scoped rule ran. Absent for a site rule (it iterates no pages) and for a `not-collected` one (it was never called, and a `0` there would read as a judgement the kernel never made). A page that emitted findings now counts as adjudicated, which it always was.

`CoverageSummary` sums them into `pageAdjudications` / `pageNotApplicable`. Named apart on purpose: those count **(rule, page) pairs**, so they happily exceed the page count, and identical names would invite reading one granularity for the other. Both spellings are documented against each other in `report.ts` and in a new `AGENTS.md` gotcha.

**Verdicts are untouched.** A rule that judged a page and found it fine still `pass`es — the count is what the report can now say alongside it. No rule decides anything differently; the fix is in what the report can express.

**Stop discarding the other pages' reasons.** `reason ??= outcome.reason` kept the first page's `notApplicableReason` and reported it as though it were the rule's. `notApplicableReasons` now carries every distinct reason in first-seen order, present iff the verdict is `not-applicable` — mirroring how `missingCapabilities` is present iff `not-collected`, so a consumer never has to know when to look. Duplicates collapse, so the uniform case (every page on movar.fyi, as it happens) stays a one-element list rather than twenty copies of the same sentence. `notApplicableReason` itself is unchanged in value and is now documented as *the first of the list, for a one-line render* — which keeps the CLI rendering added in #453 working verbatim.

Structured list rather than a joined string: a join would have changed rendered output, broken `toBe` assertions across the rule suites, and handed consumers a sentence to re-split.

## Additive, not breaking — `REPORT_SCHEMA_VERSION` stays `1`

Every new field is optional. No existing field changed type, presence, or meaning, so a stored v1 report still satisfies the `Report` type and still replays through `applySuppressions` with no shim.

The one place this could be argued the other way is `EVIDENCE_SCHEMA_VERSION`, whose doc says "bump on every additive change". That rule is stated for `Evidence`, and it earns its keep there: rules must **fork** on the version, and they do — a v1 bundle has no `head`, so the head rules answer `not-applicable` naming `schemaVersion 1` rather than passing something they never saw. Nothing must fork here. `pagesAdjudicated` absent means "not available", and every consumer's correct response to absence — old report, site rule, or `not-collected` rule alike — is to render nothing. A bump would mint a version number that nothing ever branches on, while making every downstream ask what changed in 2. The `Report`'s version is now documented as reserved for a change a reader must fork on: a field removed, retyped, or given a new meaning.

## Tests

Regression tests in `src/evaluate.test.ts`, under a new `per-page adjudication` block. All six failed first against the pre-fix kernel:

- `counts the pages a passing rule actually judged, so 1 of 3 says so` — the issue's exact case
- `counts a page that produced findings as adjudicated`
- `rolls per-page adjudication up into the coverage summary`
- `keeps every distinct not-applicable reason, not only the first page's`
- `collapses a reason every page gave alike into one`
- `summarises coverage without rolling not-collected into a pass` (existing, extended)

Two more guard the fix against over-reaching — they passed before and must keep passing: `omits the per-page counts for a site rule, which iterates no pages` and `omits the per-page counts for a rule that never ran`.

## Verification

| command | result |
| --- | --- |
| `pnpm --filter @movar/audit test` | 677 passed, 23 files |
| `pnpm nx run marketing:audit --skip-nx-cache` | `25 ran · 6 n/a · 9 not-collected`, 1 broken promise, 1 suppression applied, exit 0 — identical to `main` |

Package coverage moves up, not down: 97.26 → 97.31 % statements, 92.46 → 92.61 % branches.

## Out of scope

`collect/cli.ts` is untouched — #453 landed there mid-flight. The header still reads `coverage: 25 ran · 6 n/a · 9 not-collected`; rendering the new counts beside it is follow-up work under #448.

Closes #445


rejifald (commented):
**REVIEW: CHANGES REQUESTED — do not merge.** (GitHub refuses a formal `--request-changes` on a self-authored PR, so this is posted as a review comment; treat it as blocking.)

Reviewed adversarially against head `d4c4018`, with everything re-verified locally rather than taken from the description. **The fix itself is right, and the evidence backs every behavioural claim in the PR body.** I am requesting changes on one thing only: the four field names, which this PR's own no-bump argument makes permanent the moment it merges.

## What I verified, and what held

**The regression tests genuinely fail without the fix.** Checked out `d4c4018`, reverted *only* `src/evaluate.ts` and `src/report.ts` to `1f77e57`, kept the tests. Exactly the split claimed: 6 failed / 19 passed, and the six are precisely the six named — `counts the pages a passing rule actually judged`, `counts a page that produced findings as adjudicated`, `rolls per-page adjudication up into the coverage summary`, `keeps every distinct not-applicable reason`, `collapses a reason every page gave alike into one`, `summarises coverage without rolling not-collected into a pass`. Both over-reach guards (`omits the per-page counts for a site rule` / `for a rule that never ran`) pass pre-fix and post-fix, which is what makes them guards. Full suite on head: 677 passed / 23 files. Typecheck and lint clean.

**Nothing a rule decides changed.** Ran the CLI against `apps/marketing/dist` on the PR kernel and on `main`'s kernel from the same tree and the same build, then diffed the `--json` payloads structurally. Strip the five new keys and `collectedAt` and the two payloads are **equal including key order**. Exit `0` both. Same 40 rules, same `25 ran · 6 n/a · 9 not-collected`, same `brokenPromises: 1`, same suppression, and the CLI stdout differs only in the timestamp. `collect/cli.ts` is byte-identical to `main` and its `notApplicableReason` read is unaffected — the singular is still the first page's reason, including the `''` and no-reason fallback edges.

I also checked the merge is provably verdict-preserving rather than incidentally so: `graded.length > 0` is tested first and `graded.length` equals the total draft count, so the `adjudicated > 0` branch is reachable only when no page emitted a finding — which is exactly the old `sawPass`. `adjudicated += 1` on a `findings` outcome can therefore never flip a verdict.

**Invariants hold.** Over the full dogfood report I asserted: no `pass` with `pagesAdjudicated: 0`; no `not-collected` or site rule carrying counts; every page rule that was invoked carries both counts; `pagesAdjudicated + pagesNotApplicable == 20` for every one of them; `notApplicableReason === notApplicableReasons[0]` everywhere; no reasons list on a non-`not-applicable` result. Zero violations. A rule that adjudicated zero pages falls through to `not-applicable`, never `pass`. `evaluate.ts` remains free of `fetch`, fs, DOM, clock and randomness — the only `import` outside type-only siblings is `gradeFinding`. The empty-page edge is unreachable by construction: `deriveCapabilities` withholds `static` for a zero-page bundle, so page rules are `not-collected` before `mergeOutcomes`.

The dogfood numbers in the description are real: `core/lang-contradicts-url` is 11 judged / 9 skipped, and 394 + 186 = 580 = 29 page rules × 20 pages. `core/title-contradicts-declaration` passing on **2 of 20** pages is the single best argument for this change.

**Scope, metrics, CI.** Diff is confined to `#445`'s subject plus the metrics refresh. The README diff is one line inside the generated block (the LOC badge). `pnpm check:readme` passes on head. Package coverage measured on both sides — statements **97.27 → 97.32**, branches **92.49 → 92.64**, lines flat at 98.63 — up, not down. On the LOC jump: I ran `gen-readme-metrics.mts --refresh` on `1f77e57` and on `d4c4018` and got 56901 → 56979, i.e. **+78**, exactly this PR's net non-test TS. The committed `+262` is that +78 plus ~184 lines of pre-existing staleness inherited from `main`, since nothing gates `loc`. The refresh is accurate for the head; it is not papering over anything. Reverting the `docs/REFACTORING-QUEUE.md` churn was the right call. CI is green on `d4c4018` (`ci-gate`, `metrics-gate`, `verify`, `e2e-offline`, `build`, `audit-site`), confirmed against that headSha, and the branch is `CLEAN` against `main` under strict mode.

---

## Blocking: the four names

```json
"results":  [{ "pagesAdjudicated": 11, "pagesNotApplicable": 9 }]
"coverage":  { "pageAdjudications": 394, "pageNotApplicable": 186 }
```

`pagesNotApplicable` and `pageNotApplicable` differ by **one character**, sit in the same JSON document, hold different units, and are both optional numbers. That is not "named apart on purpose"; that is a near-homograph. `pageAdjudications` vs `pagesAdjudicated` is the same collision with the `s` moved.

Three things convince me this is a trap rather than a legible distinction:

1. **The PR has to ship a glossary entry to make the names readable.** The new `AGENTS.md` gotcha is literally titled "`pagesAdjudicated` is not `pageAdjudications`". A naming scheme that needs an "X is not Y" entry has already failed; the documentation is evidence against the names, not for them.
2. **The compiler cannot save the case that matters.** A TS consumer typing `coverage.pagesNotApplicable` gets an error, fine — but the failure this naming is meant to prevent is *reading the wrong granularity*, and `coverage.pageNotApplicable` where you wanted per-rule distinct pages type-checks perfectly. Only the name can prevent that. And the `Report` is a wire format the ADR points at Swift and Kotlin runtimes and a self-contained HTML artifact, none of which get the compiler at all.
3. **`"pageNotApplicable": 186` is not grammatical as a count.** It reads as a boolean. It is also the one name that does not follow the singular/plural convention the other three are supposed to teach.

What makes this blocking rather than a nit is **your own schema argument**. `REPORT_SCHEMA_VERSION` stays 1 because growth is additive — and by the policy this PR writes into `report.ts`, a rename is precisely "a field removed, retyped, or given a new meaning", i.e. the one thing that *does* force a bump. So this is a one-way door: cheap now, costs a version bump the moment it lands. Nothing consumes these fields yet (`#448` is still open), so now is free.

Any of these would do — the requirement is only that the coverage pair is not one `s` from the `RuleResult` pair:

- **Nest it**, so the unit comes from position and no collision is expressible: `coverage.pages: { adjudicated: 394, notApplicable: 186 }`.
- **Put the unit in the name**: `coverage.rulePagesAdjudicated` / `rulePagesNotApplicable`.
- **Rename the act**: `coverage.pageChecksRun` / `pageChecksSkipped`.

`RuleResult.pagesAdjudicated` / `pagesNotApplicable` are good names. Keep them.

## Blocking (small, and in the same edit): "every page-scoped rule that ran" is false

`report.ts` and the new `AGENTS.md` gotcha both say the coverage counts are "summed over every page-scoped rule **that ran**". `summarize` sums over *all* results, and `coverage.ran` in this very interface explicitly excludes `not-applicable`. On movar.fyi, **120 of the 186** `pageNotApplicable` pairs come from six page rules whose verdict is `not-applicable` — rules that by this module's own definition did not run. Say "every page-scoped rule the kernel invoked" (which is also the sense `pagesAdjudicated`'s doc uses, and neither is `coverage.ran`'s sense — worth disambiguating in one place given the whole point is granularity).

While in there: "Both are optional and absent means 'not available', never `0`" reads as "these are never `0`", which your own test contradicts (`pagesAdjudicated` is `0` for a rule that skipped every page, and `summarize` always emits `pageAdjudications`, including `0`). The intent — *absence must not be defaulted to zero* — is right; please say that.

---

## Not blocking, but registered

**`REPORT_SCHEMA_VERSION` staying at 1 — I accept the conclusion, but one premise is false.** The claim is that `EVIDENCE_SCHEMA_VERSION`'s bump-on-every-additive-change rule earns its keep because "rules must **fork** on the version, and they do". They do not. `page-declaration.ts:214` and `og-locale-malformed` fork on `ctx.page.document.head === undefined` — field *presence*. The version appears only inside a human-readable string (`NO_HEAD = 'this bundle predates head collection (Evidence schemaVersion 1)'`). Nothing in the package reads `evidence.schemaVersion` to branch on it. So the mechanism there is identical to the one you propose here — absence degrades — and there the project bumped anyway, with the doc saying "Bump on every additive change". Read accurately the precedent points the other way.

I still would not block, because `Report` genuinely had no stated policy before this PR, nothing breaks, no consumer must fork, and a version nobody branches on is a real cost. But two things should follow: drop the false premise, and **note the divergence at both ends** — this package will now carry two wire-format versions ~300 lines apart with opposite bump rules and nothing saying that is deliberate. One sentence on `EVIDENCE_SCHEMA_VERSION` pointing at `REPORT_SCHEMA_VERSION` would close it.

The one real cost of not bumping is worth writing down too: in a stored report, an absent `pagesAdjudicated` on a page rule with a non-`not-collected` verdict is the only signal that the report predates this change — and the new doc points readers away from that ("absent for a site rule ... and for a `not-collected` one"). Presence is a sound discriminator; it just needs saying.

**`notApplicableReasons` alongside `notApplicableReason` — sound, no objection.** It is dual carriage of one datum, which I would normally push on, but removing the singular is the single change that *would* force the version bump you are arguing against, and `#453`'s `formatResult` reads it. The two decisions are consistent with each other, the doc states the relationship exactly, and I confirmed on the dogfood report that the singular is the first element in every case and that dedup collapses 20 identical page reasons to one. Keep it.

---

Re-request review once the coverage pair is renamed and the "that ran" sentences are corrected. Everything else is ready to merge — I re-ran the whole battery above on `d4c4018` and it is otherwise clean.

rejifald (commented):
**Re-review at `514312ea`: both blocking findings are resolved. Approved on the merits.** Not merged — the branch is conflicted against a `main` that has moved four times during this review, and updating it needs a push I am not permitted to make. Details and the exact resolution I validated are at the bottom.

## 1. Naming — resolved

`CoverageSummary.rulePagesAdjudicated` / `rulePagesNotApplicable` against `RuleResult.pagesAdjudicated` / `pagesNotApplicable`. The one-character collision is gone; the `rule` prefix carries the unit. Grepped the tree: zero occurrences of `pageAdjudications` or `pageNotApplicable` anywhere, and the eleven live references are all the new names.

The part I did not ask for and think is the best thing in the rework: **"rule-pages" is now a named unit** — "one rule's look at one page" — with the reason it earns a name stated where a reader meets it (*"580 rule-pages across a 20-page site would otherwise read as an impossible number of pages"*). That is a better fix than the rename alone.

And the `AGENTS.md` gotcha I objected to is **gone**, not rewritten — the entry that existed only to tell two bad names apart is no longer needed, which is the right outcome. What replaced it earns its place: "Never default an absent count to `0`" and "'Invoked' is not `coverage.ran`" are both about the data, not about the spelling.

## 2. Doc accuracy — resolved, and I re-derived the numbers

"Summed over every page-scoped rule that ran" is gone. The replacement states the distinction explicitly and correctly: rule-pages are summed over every page rule the kernel **invoked**, which is not `ran`'s sense, and a rule invoked on 20 pages that adjudicated none contributes 20 rule-pages while counting as `notApplicable`. I re-ran the dogfood against the current base and confirmed the figure quoted in `AGENTS.md`: **120 of the 186** skipped rule-pages come from `not-applicable` rules. Exact.

The zero/absence wording is now right and matches the tests: `0` is a real value meaning the rule judged nothing, absence means nobody counted, and absence has three named causes. `expect(result.pagesAdjudicated).toBe(0)` for the skip-everything rule is consistent with all of it.

## 3. Schema rationale — the new argument survives, and it is a real one

The false premise is gone from `report.ts`, and `evidence.ts` now says plainly that the head rules fork on the field being `undefined` rather than on the version number. Correcting the sibling constant rather than quietly dropping the claim is the right shape.

The replacement argument — `Evidence` has one producer per collector runtime, `Report` has exactly one producer — **holds, and it is stronger than the one it replaces**, because it rests on the ADR's own load-bearing decision rather than on an incidental fact about rule code. I checked the asymmetry is genuine: `evaluate()` is the sole constructor of a `Report`, and `artifact.ts` and the new Safari `AuditReport.tsx` are *renderers*, not producers, so "one producer" is not quietly false now that two consumers exist. For `Evidence` the version does work that presence cannot: `DocumentEvidence.head` absent is genuinely ambiguous between "schema too old" and "this collector did not look", and only the version separates them. For `Report` there is no second writer to disagree, so presence is unambiguous. That distinction is correct and is now written down at both constants, each pointing at the other — which is what I asked for.

Stating the cost rather than burying it ("**presence is the discriminator**", with the precise rule for reading a stored report) is the part that changes my view from "defensible" to "right". I withdraw the objection.

## What I re-verified rather than assumed

- Reverted only `evaluate.ts` + `report.ts` to the merge base, kept the tests: **the same 6 fail / 2 guards pass** split holds on the reworked head.
- Full suite green at every stage: 681 on `514312ea`, 736 on the merged tree. Typecheck and lint clean on both.
- **Nothing a rule decides changed**, re-run against the new base: strip the five new keys and `collectedAt` and the `--json` payloads are equal *including key order*; `brokenPromises: 1` and exit `0` on both sides; CLI stdout differs only in the timestamp.
- Invariant sweep over the whole dogfood report: no `pass` with zero adjudicated, no counts on site or `not-collected` rules, every invoked page rule carries both, all sums equal 20, `notApplicableReason === notApplicableReasons[0]` throughout. **Zero violations.** `394 + 186 = 580 = 29 invoked page rules × 20 pages`.
- The `514312ea` LOC re-sync: recomputed authoritatively at that commit and got **57051**, matching the committed value exactly. Not papering over anything — coverage was untouched by that commit.
- CI was green on `514312ea` itself, confirmed against that headSha for both the CI and Metrics gate runs.

## What is left, and it is only mechanical

`main` landed **#462, #459, #422 and #468** while this was in review, and the branch is now behind with conflicts. `gh pr update-branch` refuses them. The conflicts are entirely in the two generated metrics artifacts — `packages/audit/AGENTS.md` auto-merges cleanly and all four families of bullets survive.

I performed and validated the resolution locally so it does not have to be rediscovered:

- Resolve `README.md` and `scripts/readme-metrics.snapshot.json` by **recomputing, not by picking a side**: `gen-readme-metrics.mts --refresh` on the merged tree gives `loc 60517`, which is main's `60398` plus this branch's own non-test TypeScript. Badge goes `60.4k → 60.5k`.
- Everything else in the snapshot stays at main's values — coverage `96.5 / 90.3`, health `82`, suppressions, bundle. I ran the full repo coverage pass on the merged tree to confirm this branch does not move the rounded figures, so `metrics-gate`'s non-waivable freshness check is satisfied by main's numbers.
- On the merged tree: 736 audit tests pass, typecheck clean, lint clean, `check:readme` clean.

Watch for one environment artifact when doing this: `pnpm install` flips `packages/audit/src/collect/cli.ts` to mode `100755` because it is a declared `bin` target. That is not a change — do not commit it.

I cannot push the update myself, so I am leaving the merge to whoever can. **Nothing about the code is blocking any more**; re-verify green on whatever head the update produces, since a green check against a stale base is not evidence and this base is moving quickly today.

rejifald:
## Review — APPROVE

Re-verified the two prior self-review rounds independently rather than taking the "resolved" declarations on trust: the near-homograph rename is complete (`git grep pageAdjudications|pageNotApplicable` → zero hits, and the glossary entry that existed only to tell the bad names apart is gone rather than rewritten); "summed over every page rule **invoked**" is now correct — I re-derived the figure and it is exactly **120 of the 186**; the `0` vs absence wording matches `expect(result.pagesAdjudicated).toBe(0)`; and the head-rule/`schemaVersion` premise is fixed (nothing in `packages/audit/src` branches on it).

### Verified

- `packages/audit`: 27 files, **899 passed**; typecheck and lint clean. `safari-host-app` (the only external `Report` consumer): 174 passed, clean.
- `nx run marketing:audit --skip-nx-cache`: exit 0, `25 ran · 6 n/a · 9 not-collected` — identical to main.
- **Negative control**: reverting only `evaluate.ts` + `report.ts` while keeping the tests fails exactly the 6 named in the body. Both over-reach guards pass in both directions, which is what makes them guards.
- **Verdict preservation proved structurally**: dumped the full `Report` for the same fixture on merged-PR and on main, stripped the five new keys — `diff` shows one line, `collectedAt`. Byte-identical including key order. `graded.length > 0` is tested before `adjudicated > 0`, so the new increment is unreachable as a verdict input.
- **Invariant sweep on the real 20-page dogfood report**: 394 + 186 = 580 = 29×20; every page rule carries both counts summing to 20; no site rule carries counts; no scored verdict with `pagesAdjudicated: 0`. Zero violations.

The premise still holds on main today: `core/title-contradicts-declaration` reports `pass` having judged **2 of 20** pages, and main's report says nothing about the other 18.

### Findings

- `packages/audit/README.md:117` + `scripts/readme-metrics.snapshot.json` — **should-fix (merge step)**: resolve by **recomputing**, not by picking a side. Measured on the merged tree: `loc: 61782` (main: 61637), badge `61.8k`. Nothing gates this — `metrics-gate.mts` has no `loc` check — so a side-pick leaves the badge wrong permanently.
- `packages/audit/src/evidence.ts:35` — **nit**: "…`schemaVersion`, which appears only inside a human-readable reason string" is stale; `artifact.ts:327` now renders `evidence v${…}` into the HTML artifact. The load-bearing half (nothing *branches* on it) is still true. Trim the trailing clause while resolving that hunk.
- `AGENTS.md:204`, `report.ts:122` — **nit**: "20 pages", "120 of the 186", "580 rule-pages" are ungated prose. All exactly correct today; nothing guards them, unlike rule counts.
- `evaluate.ts:106` — **nit**: `reasons.includes()` is O(n²) and two rules build per-page reasons, so dedup cannot collapse them on a heterogeneous site. Bounded at `DEFAULT_MAX_BUILD_PAGES = 200` (~14 KB worst case); zero rules exceeded one distinct reason on the real run. Noting the bound, not objecting.
- **PR description** — **nit**: written for `d4c4018` and never updated after the rename; it still describes `pageAdjudications`/`pageNotApplicable` and an AGENTS.md gotcha that no longer exists. Since this squash-merges, that stale text would become the permanent commit message — I will supply a corrected one at merge.

**Not superseded.** #503 reworked *finding-level* denominators; this is *page-level* coverage — a different axis. #493 makes this PR's denominator more accurate, not redundant. Main still has `let sawPass = false` / `reason ??= outcome.reason` byte-for-byte.

*Automated review.*

Issue #455, opened 2026-08-14
audit: a non-existent --dist throws ENOENT instead of exiting 2


rejifald:
`runCli(['--dist', '<path that does not exist>'])` throws an uncaught `ENOENT` out of `collectFilesystem` instead of returning the exit code `2` that `AGENTS.md` says this case should produce.

Surfaced while reviewing #450; verified against `3898c74`.

## Reproduction

```
runCli(['--dist', './definitely-not-here'], write)  ->  THREW ENOENT
```

Expected `2`. The process exits on an unhandled rejection with a stack trace rather than a diagnosable message.

## Why it matters

`packages/audit/AGENTS.md` states the contract this violates, and names this exact scenario:

> **The CLI's exit codes are three-valued.** `0` clean, `1` red (uncovered broken promises, a policy that breaks the doctrine, or a stale entry), `2` the run never happened (no source flag, an argument `parseArgs` refuses, an unreadable or unparseable policy file). A CI step that treats every non-zero as "the site is broken" will report a typo in a path as a language defect.

An unreadable *policy* file is already handled — `readPolicy` catches and returns an `Error`, and `runCli` turns that into `2`. An unreadable *dist* is not: `collect()` calls `collectFilesystem` unguarded, and `readdir` throws.

So the documented reasoning is right and one of the two paths simply was not implemented.

## Fix sketch

Mirror `readPolicy`'s shape: catch the filesystem error in `collect()` (or around it in `runCli`), and return `2` with a message naming the path — `could not read <path>`, matching the policy-file wording. The `--url` side has its own transport-error handling already and is out of scope here.

PR #464, opened 2026-08-14
fix(audit): refuse a CLI argument the parser does not recognise

### LINKED ISSUE #454: audit: an unknown CLI flag is silently ignored, so a typo audits with the wrong settings
`parseArgs` only ever looks up flags it knows by name (`argv.indexOf(flag)` over `VALUE_FLAGS`, plus `argv.includes` for the switches). It never enumerates `argv`, so any token it does not recognise is invisible — a misspelled flag is dropped in silence and the audit runs with the wrong settings.

Surfaced while fixing #431; verified against `3898c74`.

## Reproduction

```
parseArgs(['--dist', 'd', '--budgt', '5'])
  -> {"dist":"d","follow":false,"ignoreRobots":false,"ua":false}
```

The audit runs, with the default budget of 40 rather than the 5 that was asked for, and says nothing. `--folow` likewise silently disables the traversal the operator wanted.

A typo in the *source* flag is less bad but still misleading: `--dsit ./dist` exits `2` via the pre-existing "no source flag" path, so the operator gets a generic usage dump rather than "unknown flag `--dsit`".

## Why it matters

This is the same family as #431 — a typo silently changing what the audit did — but a distinct defect. #431 was a *known* flag with an unusable value; this is an *unknown* flag entirely. #431's fix (rejecting bad values, exit 2) deliberately did not add unknown-flag detection: `VALUE_FLAGS` is a closed list consulted only for value extraction.

It matters most for `--budget`, `--follow` and `--ignore-robots`, which govern how the tool behaves toward a third-party site.

## Fix sketch

After extracting known flags, walk `argv` and reject any leading-`--` token that is neither a known switch nor a known value flag (and is not the value of one). Exit `2` with the offending token named — the CLI already models `2` as "the run never happened", and `parseArgs` already returns `Args | Error` after #431.

### LINKED ISSUE #457: audit: AGENTS.md over-claims that no flag may take the next flag as its value
A gotcha bullet added in #450 says no flag may take the next flag as its value. That is true for a flag's *first* occurrence only — a repeated value flag is not checked, because `parseArgs` resolves each flag with `argv.indexOf` (first wins).

Surfaced while reviewing #450; verified against `3898c74`.

## Reproduction

```
parseArgs(['--url', 'https://x.test/', '--url', '--follow'])
  -> {"url":"https://x.test/","follow":true,"ignoreRobots":false,"ua":false}
```

No error. The second `--url` is ignored, and the `--follow` that would have been its value is picked up as a switch.

## Why it is only a doc defect

The behaviour is arguably fine: the guard fires on the occurrence whose value is actually used, and a discarded duplicate cannot change what the audit does. The problem is that `AGENTS.md` now states a broader invariant than the code holds, and `AGENTS.md` is what the next author reads before touching argument parsing.

## Fix sketch

Either narrow the sentence — say the guard applies to the occurrence a flag resolves to — or reject a repeated value flag outright and keep the sentence. The first is smaller and matches the code; the second is arguably better behaviour and would pair naturally with unknown-flag detection.

rejifald:
`parseArgs` read argv by looking each known flag up by name — `indexOf` over `VALUE_FLAGS`, `includes` for the switches — and never enumerated it. A token nobody looked for was therefore invisible:

```
parseArgs(['--dist', 'd', '--budgt', '5'])
  -> {"dist":"d","follow":false,"ignoreRobots":false,"ua":false}
```

The audit ran, with the default budget of 40 rather than the 5 that was asked for, and said nothing. `--folow` likewise left the traversal off. Both flags govern how this tool behaves toward a site that never agreed to be audited. Re-verified against `1f77e57` before fixing.

## What changed

`parseArgs` walks argv once, after the known flags are extracted, and refuses the first token that is neither a known flag nor the value of one — naming it. `runCli` already turns an `Error` from the parser into exit `2`, "the run never happened", so no exit-code plumbing was needed.

```
--dist d --budgt 5     ->  Error(unknown flag '--budgt')
--dist d --folow       ->  Error(unknown flag '--folow')
--dsit ./dist          ->  Error(unknown flag '--dsit')          # was a generic usage dump
--dist=d               ->  Error(unknown flag '--dist=d')        # see below
--dist d -budget 5     ->  Error(unexpected argument '-budget')
--dist d extra-junk    ->  Error(unexpected argument 'extra-junk')
--dist d --            ->  Error(a bare -- separates nothing: every input here is a flag)
```

**"…and is not the value of one" is the whole difficulty**, so to be explicit about the reasoning: a value flag accounts for the token after it, but **never a `--`-leading one**. No value this parser produces can begin with `--` — #450 made a flag refuse a following `--flag` as its value (`cli.ts`: `if (next === undefined || next.startsWith('--')) return new Error(...)`), precisely so `--url --follow` cannot be parsed as a url of `--follow` and then fetched. Declining to swallow a `--`-leading token here is that same rule, and it is strictly better than swallowing one: it keeps `--url a --url --folow` visible, where the duplicate otherwise discards the very occurrence that refusal was checked on.

**Positional arguments and a bare `--` — the deliberate decision.** `USAGE` documents no positional argument; every input this CLI takes is a flag or a flag's value. So a stray token is refused by the same walk rather than ignored — `movar-audit --dist d extra-junk` now exits `2` naming `extra-junk`. That closes a hole the `--`-only rule would leave open: a single-dash typo (`-budget 5`) or a shell glob that expanded would otherwise stay as silent as `--budgt` was. A bare `--` is refused on the same grounds, in its own words rather than as a misspelled flag: with nothing positional to separate, an end-of-options marker guards nothing this CLI has.

### Two behaviour notes a reader should see

- **`--dist=d` is now named rather than silently dropped.** The GNU `--opt=value` form was never supported — `indexOf('--dist')` does not match `--dist=d` — so it used to fall through to the "no source flag" usage dump, or worse, be ignored outright while some other flag carried the run. The walk sees it as one unknown token and says so. This PR does not *add* support for `--opt=value`; it stops the form failing quietly.
- **A bare `--` gets its own message**, not `unknown flag '--'`. It is not a flag anybody misspelled, so naming it one points at a fix that does not exist. What is actually wrong is that an end-of-options marker separates nothing here.

## Deliberately not fixed

Both are separate open issues against this same file, and both were checked to behave exactly as before:

- **#457** (duplicate value flags) — `parseArgs(['--url','https://x.test/','--url','--follow'])` still returns `{"url":"https://x.test/","follow":true,"ignoreRobots":false,"ua":false}`, byte-identical to the reproduction in that issue. The extraction loop was not touched, and the walk emits errors only. Its `AGENTS.md` sentence was left alone, so whoever fixes #457 still has both options that issue offers.
- **#455** (a non-existent `--dist` throws ENOENT instead of exiting 2) — nothing here inspects a path; a `--dist` value is still consumed as a value whether or not it exists.

## Tests

New in `packages/audit/src/collect/cli.test.ts` — six that fail on `1f77e57` and pass here, plus two that pass both ways and exist to catch a regression in the other direction:

| Test | Before |
| --- | --- |
| `refuses a misspelled value flag rather than auditing with the default it left in place` | fail |
| `refuses a misspelled switch rather than silently running with it off` | fail |
| `names a misspelled source flag, rather than reporting a missing source` | fail |
| `refuses a stray token, since every input this CLI takes is a flag or its value` | fail |
| `refuses a bare --, which ends an option list this CLI does not have` | fail |
| `exits 2 on an unknown flag, naming it, and audits nothing` (`runCli`) | fail |
| `accepts every flag USAGE documents, in one command line` | pass |
| `accepts the command line the dogfood gate invokes` | pass |

That last one pins `--dist dist --suppress audit-suppressions.json`, the exact shape `nx run marketing:audit` invokes — a regression there rejects the repo's own audit command and turns `audit-site` red.

## Verification

On the tree merged with `main` at `c07e215`:

- `pnpm --filter @movar/audit test` — 734 passed (24 files)
- `pnpm --filter @movar/audit typecheck` — clean
- `pnpm --filter @movar/audit lint` — clean
- `pnpm nx run marketing:audit --skip-nx-cache` — green
- `pnpm check:readme` — clean
- lefthook pre-commit ran the repo-wide `lint` / `typecheck` / `test` suite on each commit

`AGENTS.md`: the exit-code bullet gained the new invariant (appended, not reflowed). Also swept the repo for other invocation sites — the only two are the `marketing:audit` target and `docs/movar-audit-dogfood-targets.md`'s `--url <url> [--ua]`, both of which parse unchanged.

### Metrics snapshot

`ad9728e` refreshes `scripts/readme-metrics.snapshot.json` and the README badge line, which `metrics-gate` requires to be current for the merged tree. Source LOC crossed the boundary the badge rounds to: `60398` → `60456`, `60.4k` → `60.5k`. The coverage block recomputed unchanged locally. `docs/REFACTORING-QUEUE.md`, which `pnpm metrics` also rewrites, is deliberately left alone — nothing gates it and the diff is a timestamp.

Closes #454


rejifald:
Reviewed adversarially against #454 — every claim in the description re-derived rather than taken on trust. **The fix itself is correct and I would approve it on the merits.** The block is one stale number that the rebase onto current `main` produced, not a defect in the change.

## Blocking: `metrics-gate` is red on the current head

`main` moved while this sat (#462 landed), so the branch was `BEHIND` a strict-mode ruleset despite the description reporting `CLEAN`. I ran `gh pr update-branch` — clean merge, new head `dba2418` — and `metrics-gate` then failed the freshness check:

```
✗ Committed coverage is stale or inaccurate.
    committed:  96.4% lines / 90.1% branches
    recomputed: 96.4% lines / 90.2% branches
```

Reproduced locally on a tree byte-identical to the pushed head, and traced to the *interaction* with #462 — which is why neither PR could have seen it alone:

| tree | `packages/audit` branches | repo-wide | rounds to |
| --- | --- | --- | --- |
| `main` @ `228be6d` alone | 1384/1496 | 3715/4121 = 90.148 % | 90.1 ✓ |
| this PR @ `eb27284`, old base | 1392/1505 | 3723/4130 = 90.145 % | 90.1 ✓ |
| merged @ `dba2418` | 1396/1509 | 3727/4134 = **90.155 %** | **90.2** ✗ |

Each change alone lands just under the rounding boundary; together they cross it. Coverage went **up**, so the base-regression check (2) and the floor (4) are both fine — only freshness (1) is red, and freshness is deliberately not waivable by `accept-metrics-regression`. That is the right design and I am not asking for the label.

The fix is mechanical: `pnpm metrics`, commit `scripts/readme-metrics.snapshot.json` (`coverage.branches` `90.1` → `90.2`), push. The README badge renders whole percentages (`90% branches`), so it should come out byte-identical — but commit whatever `pnpm metrics` touches. Re-request and I will merge.

---

## Everything else passes

**1. Regression tests fail without the fix.** Reverted *only* `packages/audit/src/collect/cli.ts` to `1f77e57`, kept the tests: **6 failed / 672 passed**, and they are exactly the six named. The two guards pass both ways as claimed.

Both guards are load-bearing, and I would have asked for them had they been absent. The failure mode of an unknown-argument walk is over-refusal, not under-refusal, and these are the only tests on that side:

- `accepts the command line the dogfood gate invokes` pins the literal string in `apps/marketing/project.json:42`. Without it, a regression there surfaces only as a red `audit-site` job — which reads as a language defect, the exact confusion the exit-code bullet in `AGENTS.md` warns about.
- `accepts every flag USAGE documents, in one command line` is the only test that puts the full documented grammar through the walk at once; the pre-existing `parseArgs` tests each exercise one or two flags in isolation. Its `toEqual` on the whole object also catches over-consumption — a stray `index += 1` eating a token the extraction loop needed would fail here and nowhere else.

**2. Refusing positional arguments and a bare `--` — correct, and not scope creep.** This was the call I scrutinised hardest, and I land with the author.

The issue's fix sketch says "`--`-leading token". That sketch is a sketch, and taken literally it would ship a half-fix. I verified the hole directly: under a `--`-only walk, `parseArgs(['--dist','d','-budget','5'])` returns `{dist:'d',follow:false,…}` — no budget, no error, audit runs at the default 40. That is *precisely* the harm in the issue's own title, reached by dropping one keystroke instead of transposing two. Closing the two-dash spelling of the bug and leaving the one-dash spelling open is not a smaller fix, it is an incomplete one.

Nothing legitimate breaks, and I checked rather than assumed:

- `USAGE` (`cli.ts:45`) documents no positional; `Args` (`cli.ts:33`) has no positional field; `collect()` reads only `args.dist` / `args.url`. A positional is definitionally un-actionable — it can only ever have been a mistake.
- Swept `apps/`, `scripts/`, `.github/`, `docs/`, `packages/` and the root. Exactly two invocation sites exist, as claimed: `apps/marketing/project.json:42` and `docs/movar-audit-dogfood-targets.md:40`. Both are flags-only, both parse, and I ran the first one for real.
- The asymmetry favours refusing. A false refusal costs one `exit 2` with the token named; a false acceptance runs a mis-configured audit against a site that never agreed to be audited.
- Shapes that already exited `2` now exit `2` with a better message rather than a bare usage dump: `--dsit ./dist` → `unknown flag '--dsit'`. Worth noting that `--dist=d` (GNU `--opt=value`, never supported here) is likewise now named rather than silently dropped — an improvement, though the description does not mention it.

The bare `--` is the weakest of the three and deserves the caveat stated plainly: it is a real deviation from POSIX, where `--` is a harmless end-of-options marker most parsers accept and drop. I accept it because the rationale holds *given* the positional decision (with nothing to separate, `--` can only be a mistake or a wrapper artifact), because nothing in this repo injects one (no npm/pnpm script wraps the `bin` at `packages/audit/package.json:32`; nx runs the `project.json` command literally), and because the package is `private: true` and pre-publish — strict now and relaxable later is the right order for a CLI with no external callers. The `unknown flag '--'` wording is slightly awkward for a token that is not really a flag; cosmetic, not worth a round trip.

**3. The dogfood command line still parses.** `pnpm nx run marketing:audit --skip-nx-cache` green twice — on `eb27284` and again on the merged tree — one suppression applied, unchanged. Sweep results above; the "only two invocation sites" claim is accurate.

**4. Adjacent issues untouched.** Verified behaviourally, not by reading:

- #457 — `parseArgs(['--url','https://x.test/','--url','--follow'])` → `{"url":"https://x.test/","follow":true,"ignoreRobots":false,"ua":false}`, asserted **byte-identical** to the issue's reproduction by string equality, not by eye. The `AGENTS.md` sentence at line 197 ("no flag may take the next flag as its value") is unedited — it appears as a context line in the diff and greps intact on the merged tree — so both of #457's fix options stay open.
- #455 — `runCli(['--dist','./definitely-not-here-464'])` still throws `ENOENT` rather than exiting `2`. Unchanged, as claimed.

**5. Scope and invariants.** Diff is 3 files, +145/−2: `cli.ts`, `cli.test.ts`, `AGENTS.md`. None of the files in flight (`probe.ts`, `suppress.ts`, `evaluate.ts`, `report.ts`, `locator.ts`, `serving.ts`).

The three-valued exit contract holds — confirmed through `runCli`, not just `parseArgs`: unknown flag → `2`, positional → `2`, bare `--` → `2`, no source flag → `2` with the bare usage dump as before. Every pre-existing refusal returns its original message verbatim (`--url expects a value`, `--budget must be a non-negative integer, not 'lots'`), so #431's and #450's behaviour is intact. Ordering is right too: `unaccountedToken` runs after extraction but before budget validation, and `--budget -5` still earns the budget error rather than `unexpected argument`.

The `AGENTS.md` edit is accurate — I checked the one falsifiable number in it, `DEFAULT_REQUEST_BUDGET = 40` at `probe.ts:44` — and is appended to the existing bullet without reflow (one-line-per-bullet file, prettier `proseWrap` at its `preserve` default).

The `unaccountedToken` refactor is behaviour-neutral. `Error | undefined` is unambiguous — an `Error` is never `undefined`, and the helper's result is early-returned verbatim into `parseArgs`, so no behaviour can hide in the seam. `sonarjs/cognitive-complexity: 15` is enforced on `src/**` per `tooling/eslint-config-movar/configs/quality.js` and lint is clean. I also probed the final behaviour directly across ~30 argv shapes rather than reasoning from the shape of the refactor.

Local, on the merged tree: 682 tests pass (23 files), lint clean, typecheck clean, `marketing:audit` green.

Issue #439, opened 2026-08-14
audit: ua/state-language-version-lesser compares content volume across silently truncated samples


rejifald:
`contentVolume` sums `page.document.textNodes`, which the collector caps at `MAX_TEXT_NODE_SAMPLES = 1500`. Above that cap the measurement is flat, so `ua/state-language-version-lesser` compares two pages that may have wildly different real volume and sees parity — while stamping a Law 2704-VIII citation on the result.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

`packages/audit/src/rules/ua.ts:686`:

```ts
function contentVolume(page: PageEvidence): number {
  let total = 0;
  for (const node of page.document.textNodes) total += node.text.trim().length;
  return total;
}
```

Each node is additionally truncated to `MAX_TEXT_NODE_CHARS = 2000`.

## Reproduction

A 4000-node page and a 1500-node page built from the same sentence:

```
4000-node page volume: 81390 | 1500-node page volume: 81390
true node counts: 4000 vs 1500
```

Identical. `volumeFindingForPair` computes `(otherVolume - ukVolume) / otherVolume` from these numbers and fires at a 20% deficit.

## Why it matters

Two failure directions, both on a statutory rule:

- A Ukrainian page genuinely 60% shorter than its counterpart measures as parity, if both exceed the cap — the rule under-fires on the case it exists for.
- A pair straddling the cap reports a deficit that is an artifact of truncation, not of the site.

Blocked on #438: the kernel cannot even tell that truncation occurred.

## Fix sketch

Blocked on #438 (carrying the sampling report into `Evidence`). Once available, either (a) skip the pair when either side was capped and say so, or (b) compare a volume measure that survives capping. Silently comparing truncated sums is the one option that should not remain.

Issue #435, opened 2026-08-14
audit: servedLanguage classifies text the shared text-sample gates exclude


rejifald:
`servedLanguage` / `classifiedPageLanguage` hand raw `sample.text` to the classifier, skipping every gate in `text-samples.ts` — the 40-character floor, the proper-noun-run exclusion, and the `<code>`/`<samp>`/`<kbd>` exclusion. Families C, D and F go through this path; families A and E go through the gated `classifySamples`.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

`packages/audit/src/served-language.ts:63`:

```ts
const verdict = classify(sample.text, candidates);
```

versus `packages/audit/src/text-samples.ts:classifySamples`, which applies `isInsideCodeElement`, `classifiableSnippet` (which enforces `MIN_CLASSIFIABLE_CHARS = 40` and `isProperNounRun`) before classifying.

`packages/audit/src/rules/ua.ts:484` (`classifyDefaultLanguage`) has the same shape.

## Reproduction

Three samples — two two-character words and one `<code>` node:

```
classifySamples (the shared gate): 0 samples survive
servedLanguage says: {"language":"uk","via":"classified","denominator":{"examined":3,"matched":2}}
```

## Why it matters

`text-samples.ts`' module doc is explicit that this module exists to prevent exactly this: "Two copies of 'is this snippet classifiable?' would drift into two different answers to the only question that decides whether a passage gets cited in a published document about a named company."

Findings built this way carry `via: 'classified'`, so the kernel strips their failing power — no `fail` results. But they are still published observations naming a site, resting on text the catalogue's own gates say is unclassifiable. A wrong determination can also **mask** a real finding: in `core/switch-no-effect`, a bogus "the target serves a different language" verdict suppresses the draft entirely.

## Fix sketch

Route `served-language.ts` and `ua.ts:classifyDefaultLanguage` through `classifySamples`, and derive the denominator from `textNodeDenominator` so the two paths report the same measurement.

Issue #440, opened 2026-08-14
audit: x-default normalizes to a phantom language "x" in the ua pack


rejifald:
`counterpartsOf` in the `ua` pack calls `declaredLanguageOf(alternate.hreflang)` directly. `declaredLanguageOf('x-default')` returns `'x'`, so the routing declaration becomes a phantom language and the Ukrainian page is compared against the x-default target as an "x-language counterpart".

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

```
declaredLanguageOf('x-default')                        = "x"
alternateLanguage({hreflang:'x-default', …})           = null
```

`normalizeBCP47('x-default')` is `null`, so `declaredLanguageOf` falls through to `primarySubtag`, which returns `'x'`.

Every other family guards this:

- `packages/audit/src/inventory.ts:78` — `alternateLanguage` returns `null`
- `packages/audit/src/rules/switch.ts:128` — `alternateTarget` returns `null`
- `packages/audit/src/rules/inventory-hreflang.ts:144` — `hreflangGroupKey` keeps it as its own bucket

`packages/audit/src/rules/ua.ts` is the only file calling it raw — lines 192, 454 and 698. Line 698 (`counterpartsOf`) is the live one; the other two compare against `UK` so a phantom `'x'` falls through harmlessly.

## Reproduction

Actual summary emitted, with the statute citation stamped:

> `/uk/index.html` carries 0 sampled content characters against 13 for its **x** counterpart at `/index.html` (100% less). Art. 27 §6 requires the Ukrainian-language version to be no lesser in volume and content than other-language versions.

## Why it matters

`x-default` conventionally points at the default-language page — usually the largest — so the Ukrainian version gets compared against it and reported deficient, on top of whatever real counterpart comparison already ran. `inventory.ts`' doc says counting `x-default` as a language "would inflate every inventory by a phantom entry"; here it inflates a statutory finding.

## Fix sketch

Use `alternateLanguage(alternate)` from `../inventory` at `packages/audit/src/rules/ua.ts:698` and skip `null`, matching every sibling family.

PR #470, opened 2026-08-14
fix(audit): count versions, not collected pages, in the ua parity check

### LINKED ISSUE #441: audit: ua page-count parity counts collection artifacts, failing a site with exact 1:1 parity
`pageCountFindings` compares how many **collected pages** declare each language. On the most ordinary site layout — a root that duplicates the default-language version — it reports a statutory deficit on a site with exact 1:1 version parity.

## What happens

`packages/audit/src/rules/ua.ts:658` counts pages by `<html lang>` and fails when the Ukrainian count trails any other language's count.

A build of `/index.html` (a copy of `/en/`), `/en/index.html`, `/uk/index.html` — one English version, one Ukrainian version, perfect parity — counts as `en: 2, uk: 1`.

## Reproduction

```
ua/state-language-version-lesser => fail
  The site declares 1 Ukrainian-language page(s) against 2 declaring en.
  Art. 27 §6 requires the Ukrainian-language version to be no lesser in volume
  and content than other-language versions.
citation stamped: {"source":"Law 2704-VIII","article":"Art. 27 §6",
                   "url":"https://zakon.rada.gov.ua/laws/show/2704-19"}
```

The same shape occurs over the network whenever the matrix collects `/` alongside a followed `/ru/` or `/en/`.

## Why it matters

`/` and `/en/` are one version counted twice. The count is an artifact of how the collector enumerated URLs, not a property of the site — and the finding it produces cites a statute against a named company.

`ua.ts`' own module doc says the pack "reports found / not-found violation evidence" and that an undeterminable market must never `fail`. The same caution should apply to a page count that the evidence cannot actually support.

## Fix sketch

Count **versions**, not collected pages — dedupe by the hreflang cluster each page belongs to (pages that declare each other as alternates are one version), or restrict the count to pages reached as declared targets. Failing that, the rule should state what it counted and why, rather than presenting a crawl artifact as a statutory deficit.

rejifald:
`pageCountFindings` compared how many **collected pages** declare each language. On the most ordinary build there is — `/index.html` a copy of `/en/index.html`, plus `/uk/index.html`, exact 1:1 version parity — it counted `en: 2, uk: 1` and emitted a `fail` with Law 2704-VIII stamped on it:

```
ua/state-language-version-lesser => fail
  The site declares 1 Ukrainian-language page(s) against 2 declaring en.
  Art. 27 §6 requires the Ukrainian-language version to be no lesser in volume
  and content than other-language versions.
```

`/` and `/en/` are one version enumerated at two paths. The count was a property of how the collector enumerated URLs, not of the site — and it carried a statute citation against a named company.

## The design

**The two sides of the count are bounded in opposite directions, so they cannot be counted the same way.** To say the Ukrainian version is lesser you need a *lower* bound on the other language's version count and an *upper* bound on Ukrainian's.

- **Other languages: counted by version.** `versionsOf` groups collected pages by normalized location (so one URL observed from two vantages is one version observed twice) and by the hreflang alternates the pages themselves declare. `x-default` and a blank `hreflang` never merge — `alternateLanguage` is `null` for both, and since cluster membership feeds a per-language count, a routing hint would otherwise decide which language is deficient.
- **Ukrainian: counted by collected page, end to end.** Both terms — pages that serve Ukrainian, and pages that name an uncollected Ukrainian counterpart. Nothing on this side consults `versionsOf`. A partial collection cannot upper-bound how many Ukrainian pages a site has, so the most generous count the evidence permits is the only honest one.

**That is the safety property, stated exactly.** It is not "merging only joins" — false — and not "merging is barred from the Ukrainian side", which was believed once while the credit quietly iterated versions. Version-counting is safe on the side needing a lower bound and unsafe on the side needing an upper one.

Both false positives found in review were the same shape at exact parity, and both are now regression-tested: three Ukrainian pages each declaring `hreflang="uk" href="/uk/"` (the widespread "alternates name the language homepages" pattern) collapsing into one version; and then three *English* pages naming one another, collapsing three credited Ukrainian counterparts into one while the Russian count stood still.

## The credit, and its documented hole

A page naming a Ukrainian counterpart that **parses to a locator** but was never collected counts toward Ukrainian. That absence is a fact about the crawl budget or the sitemap cap, and probing the href is the one thing this pack may not do. `href="#"` and `href=""` do **not** credit — `parseLocator` answers `null` for both, and reading that as "exists but uncollected" let markup asserting nothing silence a statutory check.

Known and accepted, documented in `ua.ts` and `AGENTS.md` rather than papered over: a site can suppress this half with one fabricated `uk` alternate per page, and on an ordinary network run **nothing reports it** — `core/hreflang-target-unresolvable` needs `traversal`, which network evidence carries only when some page was reached as a declared target. It is caught off disk. The cost is accepted on the pack's standing bias: a missed violation is recoverable, a statute-citing accusation against a compliant site is not. `ua/state-language-absent` already carries the identical hole.

Rejected: **gating the credit on `traversal`** (it withholds the credit exactly where the crawl-truncation false positive lives, trading a missed violation for a false one) and **requiring same-origin** (it does not close the hatch — the same-origin dangling href still credits — and would false-fail a legitimate Ukrainian version on a separate domain). Making a credited-but-unverifiable claim *visible* in the report is the real answer, and is being filed separately.

## Tests — what fails against what

Reverting only `packages/audit/src/rules/ua.ts` and running `src/rules/ua.test.ts`:

| baseline | failures |
| --- | --- |
| `origin/main` | **26 of 100** |
| round-1 head `4cdd66f` | **15** |
| round-2 head `4bf0cf2` | **6** |

The layered baselines matter, and an earlier revision of this description got it wrong: **not every new test is a regression against `main`.** The two merge-guard tests (`x-default` on the Ukrainian pages; Ukrainian alternates naming the Ukrainian homepage) *pass* on `main` — those sites were never broken there. They guard the round-1 head, which is what introduced the defect they describe. Likewise the credited-basis and property tests guard round-2. All are load-bearing; only some are regressions against `main`.

The headline property, which would have caught every false positive this rule has had:

> **adding an hreflang alternate to pages that do not serve Ukrainian never turns a `pass` into a `fail`.**

It mutates **one `<html lang>` group at a time** on purpose — mutating every non-Ukrainian page at once collapses both sides symmetrically and hides the defect, which is exactly what happened to this test's first draft (it passed against the broken head). Against round-2 it now fails four cases, covering the credited-counterpart and `lang`-less shapes.

Existing coverage is intact: the 3-versus-1 deficit, the absolute `0 collected page(s) declaring Ukrainian` deficit, and the parity pass all still fire.

## Verification

- `pnpm --filter @movar/audit test` — 778 passed / 25 files (`ua.test.ts`: 100)
- `typecheck`, `lint` — clean
- coverage — `ua.ts` 100% stmts/funcs/lines, 98.87% branch; the two uncovered branches are pre-existing and untouched
- `pnpm nx run marketing:audit --skip-nx-cache` — succeeds; `ruleset movar-audit-core@0.0.0`, `grep -c "ua/"` = **0**, so the pack is genuinely not composed there
- metrics regenerated with `apps/marketing/.astro` deleted first, so no #472 inflation: `loc` 60347 → 60549, exactly the net +202 source lines this branch adds to `ua.ts` (tests are excluded from the walk)

## Scope

`volumeFindingForPair` / `contentVolume` untouched — the truncated-sample problem there is #439. `classifyDefaultLanguage` untouched — #435.

Closes #441


rejifald (commented):
# CHANGES REQUESTED — do not merge

Reviewed adversarially at head `4cdd66f`, base `228be6d`, against `main` at `3c09d02`. The bug in #441 is real and the location/cluster dedupe is the right shape for it. Four of the seven checks pass. Three do not, and two of those are the kind this pack exists to prevent: a **new** false accusation carrying Law 2704-VIII, and an escape hatch that nothing else catches.

## 1. Regression tests fail without the fix — PASS

Reverted only `packages/audit/src/rules/ua.ts` to `origin/main`, kept the tests:

```
Tests  6 failed | 671 passed (677)
  fires on a sitewide page-count deficit
  does not count a page with no determinable served language toward either side …
  fires an absolute deficit when the market is determined but the site declares no Ukrainian pages at all
  does not count a root page and its language-specific twin as two versions of one language
  counts one URL observed from two vantages as one version, not one per observation
  does not count a version against Ukrainian when it declares a Ukrainian counterpart the run never collected
```

Exactly the 3 new tests plus the 3 wording assertions, as stated. On the head: 677 passed (23 files), `typecheck` clean, `lint` clean. Confirmed.

## 2. The credit — FAIL. It is an escape hatch, and nothing catches the escape

Built the adversarial site and measured through `evaluate` with `withPack(CORE_RULESET, ...UA_PACK_FAMILIES)`. Five `.ua` pages, `<html lang="ru">`, zero Ukrainian pages anywhere:

| site | `ua/state-language-version-lesser` | `core/hreflang-target-unresolvable` |
| --- | --- | --- |
| no hreflang at all (baseline) | **fail** (5 ru vs 0 uk) | not-collected |
| `hreflang="uk" href="/uk/<slug>"`, never collected | **pass** | **not-collected** |
| …same, with probes attached | **pass** | **not-collected** |
| `hreflang="uk" href="#"` | **pass** | **not-collected** |
| `hreflang="uk" href=""` | **pass** | **not-collected** |
| `hreflang="uk" href="https://not-our-site.example/uk/"` | **pass** | **not-collected** |
| …same dangling `uk`, but off disk | **pass** | fail ×5 |

Two things fall out of that table.

**(a) The backstop is not there on a network run.** `core/hreflang-target-unresolvable` declares `capabilities: TRAVERSAL_ONLY`, and on network evidence `hasTraversal` is `pages.some(page => page.reach === 'declared-target')`. A site whose declared `uk` targets do not exist produces no page with that reach, so on the ordinary sitemap/entry-point crawl the capability is absent and the rule reads `not-collected` — never a finding. The dangling target is caught *only* off disk, where traversal is free. So the pack, composed as a caller composes it, goes completely silent on a site that emits one fake `uk` alternate per page. The asymmetry argument is sound in principle; it is being applied where the "someone else catches it" half is not actually true.

**(b) The credit is granted on markup that declares nothing.** `href="#"` and `href=""` both credit Ukrainian. `parseLocator` returns `null` for an empty href and for a bare fragment — the locator module's own doc says "`#uk` declares no target" — so `resolveTargetPage` returns `null`, and `declaresUncollectedUkrainian` reads that `null` as "the counterpart exists but we did not fetch it". That is not the rule the PR describes. The stated principle is that the site's own markup *asserts the version exists*; `hreflang="uk" href="#"` asserts nothing at all, and it silences a statutory check. Minimum fix: require the href to parse to a locator before crediting.

**Verdict on the credit's merits.** The direction is defensible and the "observed against / declared for" asymmetry is consistent with `ua/state-language-absent`, which already passes on a declared-but-unverified `uk` alternate. The uncollected-but-parseable same-origin case (crawl budget) is a reasonable thing to credit. What is not defensible as written is crediting a non-target, and leaning on a backstop that is capability-gated out of existence in the exact configuration where the escape works. Either narrow the credit (parseable locator, and consider requiring same-origin) and say in the module doc that on a `traversal`-less network run nothing reports the dangling href, or gate the credit on `traversal` being present so the run that cannot check is the run that does not credit.

**(c) The summary misstates its own basis.** A version credited through a dangling href is reported as "declaring Ukrainian". One `.ua` site, five `ru` pages, one dangling `uk` alternate on `/ru/` only:

```
The 5 collected pages resolve to 5 distinct version(s) declaring ru against 1 declaring Ukrainian …
```

Zero collected pages declare Ukrainian. The finding tells a named company that one version declares Ukrainian when what happened is that one version named an uncollected counterpart. Since the whole point of the new wording is that "a finding that does reach a site owner names its own basis", the credited versions need to be named as credited.

## 3. `x-default` in the clustering — FAIL, and it is not load-bearing

The `alternateLanguage` null-for-`x-default` contract is respected wherever a *language* is decided: `versionLanguages` reads language only from `declaredServedLanguage` (`<html lang>`), and `declaresUncollectedUkrainian` filters on `alternateLanguage(alternate) === UK`, so an `x-default` never credits Ukrainian. #440's letter holds.

Its spirit does not. Cluster membership feeds straight into a per-language count, so an `x-default` now decides which language is deficient. Measured — three `uk` pages that declare `x-default → /uk/`, three `ru` pages that declare nothing:

```
main:      pass
this PR:   fail — "3 distinct version(s) declaring ru against 1 declaring Ukrainian"
```

A site with exact 3-versus-3 page parity is newly accused under Law 2704-VIII because its Ukrainian pages carry an `x-default` and its Russian pages do not.

And the inclusion buys nothing. Patching `mergeDeclaredCluster` to skip `x-default`:

- the whole `ua.test.ts` suite still passes, **62/62**, including this PR's own issue-repro test;
- the over-fire above becomes `pass` again.

The issue's build never needed it: the root already declares `hreflang="en" → /en/index.html`, which unifies `/` with `/en/` on its own. Drop `x-default` from the merge.

## 4. The three changed assertions — PASS

All three are genuine wording updates. Every number is preserved (`1`/`3`, `1`/`3`, `0`/`2`), `toHaveLength(1)`, the verdict assertion and the citation assertion are all retained, and no assertion was deleted. The three the PR says still fire do fire, on the head: the 3-versus-1 deficit, the absolute `0 declaring Ukrainian` deficit, and the parity pass — the last is untouched and still does real work (`/uk/` merges with `/ru/`, `/uk/about` and `/ru/about` stay separate, 2 versus 2).

## 5. Scope — PASS

`volumeFindingForPair`, `contentVolume` and `classifyDefaultLanguage` do not appear anywhere in the diff. `AGENTS.md` is a clean single-line append (`@@ -200,3 +200,4 @@`, one `+`, zero `-`) with no reflow. Nothing under `packages/audit/src/` outside `rules/ua.ts` and its test.

**But the appended invariant is false**, and that is worse than a code bug because it is written down as doctrine for whoever changes this next:

> Merging on declarations can only join versions, never split them, so sloppy hreflang costs count coverage rather than manufacturing a deficit.

Joining does not have to split anything to manufacture a deficit — it only has to fall on the Ukrainian side. Besides the `x-default` case above, the plain-`hreflang` version: three `uk` pages each declaring `hreflang="uk" href="/uk/"` (the widespread "alternates name the language homepages" pattern), three `ru` pages declaring nothing → `main: pass`, this PR: **fail**, 3 ru versions against 1 uk. Merging is only safe in the direction claimed when it is *symmetric across languages*, and nothing here makes it so. Please fix the invariant along with the code — the safety argument for the whole change rests on it.

## 6. The dogfood gate — PASS, verified rather than assumed

`pnpm nx run marketing:audit --skip-nx-cache` succeeds. `ruleset movar-audit-core@0.0.0`, `grep -c "ua/"` on the full output is **0**. 23 pass / 6 not-applicable / 9 not-collected / 1 suppressed, unchanged.

## 7. CI green on a current head — FAIL, and the snapshot in the diff is wrong

Head `4cdd66f` has **0 check runs**; `gh pr checks 470` reports none. `mergeStateStatus: DIRTY`. `git merge-tree` confirms the conflict is confined to `README.md` and `scripts/readme-metrics.snapshot.json`; `packages/audit/AGENTS.md` auto-merges and `packages/audit/**` is untouched by `main`, so the substance merges cleanly (#471).

I did not update the branch and re-run `pnpm metrics`, because the code has to change for the items above and the snapshot would only have to be redone. When it is redone, note that the committed value is currently inflated by #472. Replaying the generator's own walk over `apps/` + `packages/`:

```
with apps/marketing/.astro present : loc = 57049   ← the value committed here
with apps/marketing/.astro absent  : loc = 56839
.astro's contribution              : 210 generated lines
```

`57049` matches to the digit, so `pnpm metrics` was run over a present build directory. The README line in this diff should read `56.8k`, not `57.0k`. Delete `apps/marketing/.astro` before re-running.

---

## What would unblock this

1. Do not credit an alternate whose href does not parse to a locator (`href="#"`, `href=""`).
2. Close the network-run hole: either gate the credit on `traversal`, or state plainly in the module doc that `core/hreflang-target-unresolvable` is `not-collected` on a run with no declared-target reach, so the dangling href goes unreported by design.
3. Name credited versions as credited in the summary, instead of reporting them as "declaring Ukrainian".
4. Drop `x-default` from `mergeDeclaredCluster` — 62/62 still pass without it, and it is the sole cause of one of the two new false positives.
5. Fix the plain-`hreflang` same-language merge, or bound the merge so it cannot reduce one language's count without reducing the others'.
6. Correct the `AGENTS.md` invariant so it says what merging actually guarantees.
7. Update the branch, delete `apps/marketing/.astro`, re-run `pnpm metrics`, and get a full green run.

Everything above was measured on this head, not read off the diff. Happy to re-review as soon as it moves.

rejifald (commented):
# CHANGES REQUESTED — do not merge

Re-reviewed at head `4bf0cf2` against `main` at `47df24f`. Five of my six round-1 items are genuinely closed, and the reformulation of the safety property is the right one — it is the formulation I would have asked for. **But the code does not yet implement it.** One blocker remains, it is one line, and I have verified the fix.

## The asymmetry claim — the principle is right; the implementation contradicts it

The claim, as written into `ua.ts` and `AGENTS.md`:

> Ukrainian is counted by **collected page**, never merged, plus versions naming a Ukrainian counterpart the run did not collect. […] Merging is not safe because it only joins; it is safe only because it is barred from the Ukrainian side.

I accept that framing. Lower bound on the other side, upper bound on Ukrainian's, so the two sides cannot be counted the same way — that is correct, and it is a better answer than the one I asked for.

The problem is the second clause. **`credited` is version-counted**, so merging is *not* barred from the Ukrainian side:

```ts
for (const version of versions) {
  if (version.some((page) => servesUkrainian(page))) continue;
  if (version.some((page) => declaresUncollectedUkrainian(page, pages))) credited += 1;
}
```

`served` is page-counted and safe. `credited` iterates `versions`, so every merge that joins two credited versions removes a unit from the Ukrainian side. When the merge lands on pages that serve some *other* language, the other-language count does not fall with it — and the deficit appears out of nothing.

Three sites, measured through `evaluate` with `withPack(CORE_RULESET, ...UA_PACK_FAMILIES)`. In each pair the site substance is identical; only hreflang sloppiness on the **non-Ukrainian** pages differs:

**A/B — the cleanest one.** 3 Russian pages, 3 English pages, and 3 Ukrainian counterparts the run never collected but which every English page names. The Ukrainian side is at exact parity with Russian: 3.

| | English pages cross-declare? | verdict |
| --- | --- | --- |
| A | no | **pass** |
| B | yes — `hreflang="en" href="/en/1"` | **fail** — "3 version(s) declaring ru … against 0 collected page(s) declaring Ukrainian and 1 version(s) naming a Ukrainian counterpart this run did not collect" |

The English pages collapse 3 versions into 1, and drag the Ukrainian credit 3 → 1 with them. The site is accused under Law 2704-VIII because *its English pages* use the "alternates name the language homepages" pattern — the exact pattern the new doctrine cites as the reason not to merge the Ukrainian side. It reaches the Ukrainian side anyway, through the credit.

**C/D.** Three credited versions serving different languages (`en`, `de`, `fr`), 3 Russian pages. Unmerged: **pass**. Joined by their own hreflang: **fail**. Merging two credited versions that serve *different* languages costs the Ukrainian side one unit and costs no other-language count anything.

**E/F.** Same shape with two `<html lang>`-less credited pages: unmerged **pass**, merged **fail**.

So the safety property holds for `served` and fails for `credited`. As it stands the rule's own comment is falsified by the code four lines below it.

### The fix, verified

Count the credit per page, exactly as `served` is counted:

```ts
const credited = pages.filter(
  (page) => !servesUkrainian(page) && declaresUncollectedUkrainian(page, pages),
).length;
```

Measured on this head with that substitution:

- A/B, C/D, E/F — **all six now pass**; merging can no longer move the Ukrainian side at all;
- `pnpm --filter @movar/audit exec vitest run src/rules/ua.test.ts` — **67/67 pass**, including this PR's own new credited-basis test and both new merge-guard tests;
- my round-1 counter-examples stay fixed, and the `href="#"` / `href=""` sites stay `fail`.

It costs nothing and it makes the sentence in `AGENTS.md` literally true: nothing on the Ukrainian side is merged. Two follow-ons, both small: `describeUkrainianCoverage` should then say `page(s)` rather than `version(s)`, and the doctrine should name *what* is barred — "the Ukrainian side is page-counted end to end, `served` and `credited` alike" — because "barred from the Ukrainian side" is the claim that was already believed once while `credited` quietly wasn't.

A note on why this keeps happening in the same spot: version-counting is safe on the side you need a *lower* bound for and unsafe on the side you need an *upper* bound for. `credited` sits on the upper-bound side but was written in the lower-bound idiom. A short property test — "adding any hreflang alternate to a non-Ukrainian page never turns a pass into a fail" — would have caught all three of my constructions and both of my round-1 ones, and would catch the next one. Worth considering; not a condition of merge.

## The documented escape hatch — ACCEPTED

You asked me to rule. **A documented, unreported escape hatch is acceptable here.** I withdraw the round-1 objection.

- Gating on `traversal` is worse, and for the reason you give: on a network run without traversal it would withhold the credit entirely, so every site whose Ukrainian pages merely fell outside the crawl budget gets a statute-citing accusation. That trades a missed violation for a false one, against the pack's standing bias.
- Declining the same-origin restriction is right. It does not close the hatch — the same-origin dangling href still credits — and it would false-fail a legitimate Ukrainian version on a separate domain.
- `ua/state-language-absent` already carries the identical hole (`declaresUkrainianVersion` → `pass`), so closing it in this one rule buys a site nothing. It is a pack-level question, not this PR's.
- The `parseLocator` gate closes the only part that was indefensible. Verified: `href="#"` and `href=""` now both `fail`; uncollected same-origin and off-site parseable hrefs still `pass`, as documented.

What would actually close it, for whoever picks it up: not a stricter credit, but a *visible* one — the run should say that a credited claim was unverifiable, rather than falling silent. Either an `info`-graded ua finding naming the credited hrefs, or letting `core/hreflang-target-unresolvable` distinguish "not checked" from `not-collected`. Please file that; it should not hold this PR.

The documentation in `AGENTS.md` names the mechanism precisely — `traversal`, `reach === 'declared-target'`, and where it *is* caught. That is what I asked for.

## The rest — all verified

**`x-default` dropped — confirmed, and it cost nothing.** `mergeDeclaredCluster` now skips anything `alternateLanguage` calls `null`. My round-1 sites both pass: `x-default` on the Ukrainian pages only, and `hreflang="uk"` naming the Ukrainian homepage — 3-versus-3 parity, `pass`. Full suite green: **745 passed (25 files)**.

**Regression tests — verified, with one correction to the record.** The three original tests are byte-identical to the round-1 head. Reverting only `ua.ts` to `origin/main` gives 9 failures — the three originals, the three wording assertions, the `it.each` pair, and the credited-basis test. But the two new merge-guard tests (`x-default` / Ukrainian-homepage alternates) **do not** fail against `main`; those sites pass there. They fail against the round-1 head (`4cdd66f`), which is what they are guarding. Genuine and load-bearing tests — just not regressions against `main`, so the blanket "all fail without the fix" is not right.

**Wording assertions.** Third revision, numbers preserved again (1/3, 1/3, 0/2), and `0 collected page(s) declaring Ukrainian` is more precise than either predecessor. The credited-basis summary no longer merges credited versions into "declaring Ukrainian" — my round-1 (c) is closed.

**Scope.** Still exactly five files. `AGENTS.md` is a clean two-line append at line 227, no reflow. `volumeFindingForPair`, `contentVolume`, `classifyDefaultLanguage` absent from the diff.

**Dogfood gate.** `pnpm nx run marketing:audit --skip-nx-cache` succeeds, `ruleset movar-audit-core@0.0.0`, `grep -c "ua/"` = **0**.

**Metrics.** `loc 60434` matches the generator's own walk with `apps/marketing/.astro` absent, to the digit. The #472 inflation is out. Good.

**CI.** Still **0 check runs** on `4bf0cf2`; `gh pr checks` reports none; `mergeStateStatus: DIRTY`. `git merge-tree` now shows a single conflicted file, `scripts/readme-metrics.snapshot.json` — `README.md` and `packages/audit/AGENTS.md` both auto-merge. I did not run the update-branch/metrics step, since the code has to move again and the snapshot would only have to be redone. Your plan to leave it as the immediately-before-merge step is right, and I will do it then.

---

## To unblock

1. Count `credited` per page, not per version.
2. `describeUkrainianCoverage` → `page(s)`; adjust the credited-basis test's expected string.
3. Reword the invariant to say the Ukrainian side is page-counted end to end, `served` and `credited` alike.
4. Then `update-branch`, delete `apps/marketing/.astro`, `pnpm metrics`, green run.

Optional, and I would not hold the PR for it: the property test above, and an issue for making a credited-but-unverifiable claim visible in the report.

Everything above was measured on `4bf0cf2`. The design is right now; it is the last four lines of arithmetic that aren't. Ping me and I will re-run.

Issue #438, opened 2026-08-14
audit: the digest's sampling report never reaches Evidence, so truncation is invisible to every denominator


rejifald:
`digestDocument` computes a `SamplingReport` recording that the 1500-node cap was applied, and the collector then drops it: `PageEvidence` has no field for it. Nothing downstream can tell that a page's text samples were truncated, so every "N of M sampled text nodes" denominator understates M.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

`packages/audit/src/collect/digest.ts:418` builds the report; `packages/audit/src/collect/node.ts:150` keeps it only on the internal `CollectedPage`, and `collectFilesystem` destructures `const { document } = digestDocument(html, {})`, discarding it. `PageEvidence` is `{ id, url?, path?, reach, rendered, document }` — no sampling field. (The `cappedAt` at `packages/audit/src/evidence.ts:262` belongs to `ScreenshotViewport` and is unrelated.)

## Reproduction

A page with 4000 eligible text nodes:

```
digest sampling report: {"examined":4000,"sampled":1500,"cappedAt":1500}
textNodes that survive into Evidence: 1500
serialized page has a sampling field? false | in document? false
denominator the report will quote: {"examined":1500,"matched":3}
true eligible nodes on that page: 4000
```

## Why it matters

`digest.ts`' own doc on `MAX_TEXT_NODE_SAMPLES`: "When it bites, the caller is told — see `DigestResult.sampling` — because a silently truncated sample makes every denominator downstream a lie." The telling stops at the collector boundary.

`finding.ts` calls the denominator the difference between a finding and a smear: *"3 of 340 text nodes" is a finding; "3 nodes" is a smear.* Understating the denominator by 2.7× **inflates** the apparent share in every classified summary and in `core/content-contradicts-declaration`'s wording.

## Fix sketch

Add the sampling report (or at minimum `examined` and an optional `cappedAt`) to `DocumentEvidence`, populate it in both collectors, and have `textNodeDenominator` report the true examined count. `schemaVersion` exists for exactly this kind of additive growth.

Issue #472, opened 2026-08-14
tooling: the README LOC badge counts generated .astro files, so it depends on whether a build has run


rejifald:
`BUILD_DIRS` in `scripts/gen-readme-metrics.mts` does not list `.astro`, so the LOC walk descends into `apps/marketing/.astro/` and counts generated, gitignored TypeScript as source. The README's LOC badge therefore depends on whether `astro sync` has run in the working tree.

Found while reviewing #461; verified against `9cf8a38`.

## The omission

`scripts/gen-readme-metrics.mts:199`:

```ts
const BUILD_DIRS = new Set([
  'node_modules', '.output', 'dist', '.nx', '.wxt', 'coverage', '.turbo',
]);
```

`.astro` is absent, and `.gitignore:8` ignores it — so the directory is generated output by the repo's own definition, and the walk at line 222 recurses into it, counting every `.tsx?` it finds.

## Measured

| tree state | `loc` |
| --- | --- |
| clean checkout at the reviewed head | 56712 |
| same head after `astro sync` | **56922** |

The +210 is exactly `apps/marketing/.astro/content.d.ts` (208) plus `types.d.ts` (2). The badge renders `56.9k` where the truth is `56.7k`.

## Why it matters beyond two hundred lines

- **The value is non-deterministic.** Whether the badge is right depends on whether a contributor happened to run a build before `pnpm metrics`. Two people on the same commit legitimately produce different numbers.
- **It has been drifting for a while.** The same artefact inflated the count at #406, #424 and #427 (+206 / +469 / +176), so the committed figure has been wrong across several releases.
- **It corrupts a published claim.** The README badge is a stated repo metric, and `README.md` parity is itself gated by `pnpm check:readme` — which compares the README against the snapshot, not the snapshot against reality, so it cannot catch this.

It also muddied a review: #461's author attributed their `loc` delta to drift from other PRs, when part of it was this artefact growing. Of a committed +205, roughly +132 was genuine drift, +39 the PR's own source, and +34 growth of the generated file.

## Fix sketch

Add `.astro` to `BUILD_DIRS`. Better, derive the exclusion set from `.gitignore` rather than maintaining a parallel hand-written list that silently diverges every time a tool adds a cache directory — `.astro` is the second such omission this list has had to learn.

Worth pairing with #471, which tracks the wider problem that nothing checks `loc` against reality at all.

rejifald:
Confirmed live on `main`, twice, by two independent reviews.

[...]

PR #478, opened 2026-08-14
fix(audit): refuse the ua volume comparison across a truncated sample

### LINKED ISSUE #439: audit: ua/state-language-version-lesser compares content volume across silently truncated samples

rejifald:
Closes #439

## The defect

`ua/state-language-version-lesser` compares per-page content volume across
hreflang pairs by summing `page.document.textNodes` — a sample the collector
caps at `MAX_TEXT_NODE_SAMPLES` (1500). Above that cap the sum is flat, so the
rule compared two floors and published the result as a measurement, with Law
2704-VIII Art. 27 §6 cited on it.

Both directions were wrong:

- **Under-fires on the case it exists for.** Two capped pages of wildly
  different real size measure identically, so a Ukrainian version genuinely a
  third the size of its counterpart reads as exact parity.
- **Over-fires on the collector.** A pair straddling the cap publishes a deficit
  that belongs to the sampling ceiling rather than to the site — a false
  accusation, statute cited, against a named company.

Unblocked by #438 (`47df24f`), which put the digest's sampling report on the
wire as `DocumentEvidence.textSampling`. `evidence.ts` already documents
`cappedAt`'s presence as "how a rule asks *may I still compare these pages by
volume?*" — this is the rule that had to ask.

## Before / after

Same evidence: a `/uk/` page whose 9000 examined text nodes were sampled down
to 1, against an untruncated `/ru/` counterpart.

**Before** — a `fail` finding:

> https://example.com.ua/uk/ carries 15 sampled content characters against 181
> for its ru counterpart at https://example.com.ua/ru/ (92% less). Art. 27 §6
> requires the Ukrainian-language version to be no lesser in volume and content
> than other-language versions.

**After** — a `declared`/`info` finding, no `fail`:

> Content volume was not measured between https://example.com.ua/uk/ and its ru
> counterpart at https://example.com.ua/ru/: the collector's text-sampling cap
> truncated https://example.com.ua/uk/ (1 of 9000 examined text nodes), so
> comparing their sampled character counts would measure that cap rather than
> these pages. Art. 27 §6 requires the Ukrainian-language version to be no
> lesser in volume and content than other-language versions; this pair was not
> assessed against that requirement.

## The fix

`volumeFindingForPair` gates on the new `truncationDetail(page, locator)` before
computing any volume, and refuses the pair when either side was truncated.

Three deliberate choices:

- **Keyed on the presence of `textSampling.cappedAt`** — not on `examined >
  sampled`, and not on the sample length measured against a hard-coded
  `MAX_TEXT_NODE_SAMPLES`. That constant is the collector's, and the kernel
  adjudicates bundles from collectors it has never seen.
- **A pre-v3 bundle carries no report and is read as untruncated.** That is
  exactly how it was always compared; reading an unknown as a truncation would
  silence the rule on every stored bundle, including the small pages the cap
  never came near.
- **`info`, not silence.** Suppressing the finding would leave the rule
  reporting `pass` on a pair it never measured — indistinguishable from one it
  measured and found clean. That is the same failure mode `not-collected is
  never pass` exists to prevent. `info` findings do not count toward
  `brokenPromises`, so a collector artifact still cannot turn a build red.

The summary is true on both capped and uncapped pages: it names only the sides
whose evidence carries `cappedAt`, quotes their sampled-of-examined counts, and
makes no claim about a deficit it did not measure.

## How the regression tests prove it

Five cases under `ua/state-language-version-lesser > a truncated text sample` in
`packages/audit/src/rules/ua.test.ts`, driven through `evaluate` with builders
from `test/fixtures.ts`, vitest `globals: false`:

| Test | Guards |
| --- | --- |
| `does not accuse a Ukrainian page whose own sample was the one truncated` | the false-accusation direction |
| `does not read parity from two samples that both hit the cap` | the under-firing direction — the issue's own reproduction |
| `does not accuse a Ukrainian page whose counterpart was the truncated one` | the same refusal in the other orientation |
| `still compares two pages whose samples were not truncated` | the gate keys on `cappedAt`, not on the report's presence |
| `still compares a bundle collected before the sampling report existed` | pre-v3 bundles compare as they always did |

Verified by reverting **only** `src/rules/ua.ts` to `origin/main` and keeping the
test file — the first three fail, exactly on the defect:

```
× does not accuse a Ukrainian page whose own sample was the one truncated
    AssertionError: expected [ { …(8) } ] to deeply equal []
    + "summary": "https://example.com.ua/uk/ carries 15 sampled content characters
    +   against 181 for its ru counterpart at https://example.com.ua/ru/ (92% less).
    +   Art. 27 §6 requires the Ukrainian-language version to be no lesser in volume
    +   and content than other-language versions.",
    + "verdict": "fail",

× does not read parity from two samples that both hit the cap
    AssertionError: expected [] to have a length of 1 but got +0

× does not accuse a Ukrainian page whose counterpart was the truncated one
    AssertionError: expected [ { …(8) } ] to deeply equal []

Tests  3 failed | 61 passed (64)
```

The last two pass in both states — they exist to catch the gate over-firing.

## Verification

- `pnpm --filter @movar/audit test` — 25 files, 755 tests pass; lint and
  typecheck clean.
- `pnpm validate` — exit 0. typecheck 21 projects, lint 21 projects, test 19
  projects, and `audit:site` (the dogfood gate) all green. The `ua` pack is
  deliberately not composed into the dogfood ruleset, so no suppression became
  stale and none was added.
- `pnpm test:coverage && pnpm gen:readme --refresh` — coverage **unchanged**
  (lines 96.6, branches 90.3); `loc` 60347 → 60616 from the new source. `astro
  sync` was run first, so the `.astro/*.d.ts` files are present and the `loc`
  figure is not the bogus-decrease case. Snapshot committed. No
  `accept-metrics-regression` label.

  (Figures refreshed after the rebase onto `9a5e5cd`. An earlier revision of
  this description quoted the pre-rebase run — 748 tests, `loc` 60518 → 60577,
  coverage 96.5 → 96.6 — which was measured against a different base.)

## Out of scope

`classifyDefaultLanguage` in the same file builds its `Denominator` from the
filtered `textNodes.length` instead of calling `textNodeDenominator`, so
`ua/state-language-not-default` still quotes the cap rather than the population
on a truncated page — overstating the published share in the direction of the
accusation, on another statute-cited rule. Untouched here: different rule,
different mechanism, and coherence does not require it. Worth its own issue.

## Documentation

- `packages/audit/AGENTS.md` — the `textNodes` is a sample gotcha now records
  that the flag, not `examined > sampled` and not a hard-coded ceiling, is how
  to ask, and what this rule publishes when it refuses.
- `docs/movar-audit-rules.md` — the family F catalogue entry records the refusal
  and the `info` finding.


rejifald:
## Review gate: **BLOCKED** — on one mechanical item, not on the change itself

The fix verifies clean on every substantive axis. The only blocker is that `main` moved while this
was under review: **#470 (`d50d1c3`) landed at 02:18** and this PR now reads `CONFLICTING` / `DIRTY`.

### The blocker

`git merge-tree origin/main 933613d` conflicts in exactly two files, both generated:

```
CONFLICT (content): Merge conflict in README.md
CONFLICT (content): Merge conflict in scripts/readme-metrics.snapshot.json
```

`packages/audit/src/rules/ua.ts`, `ua.test.ts` and `AGENTS.md` **auto-merge cleanly** — #470 rewrote
`pageCountFindings` and explicitly left `volumeFindingForPair` / `contentVolume` alone for this issue,
so the two changes touch disjoint regions.

I resolved the merge locally and confirmed the two changes **compose semantically**:

```
src/rules/ua.test.ts       105 passed (105)
@movar/audit (full)  25 files, 796 passed (796)
```

So this is a metrics-snapshot refresh, not a code conflict.

**One wrinkle to decide before regenerating.** #470 also changed the `loc` methodology — it refreshed
the snapshot with `apps/marketing/.astro` *absent*, so the figure is no longer inflated by generated
build output (#472). This PR's snapshot (`loc` 60616) was generated the other way, with `astro sync`
run first so the `.astro/*.d.ts` files were present. `main` is now `loc` 60549 on the new convention.
Regenerate on the merged tree under #470's convention (no `astro sync` first), or the `loc` figure
will carry the old methodology forward. `metrics-gate` is non-waivable, so it will have the last word
either way.

After rebasing: re-run all six checks, and the Verification section of the description will need its
`loc` figure refreshed once more.

### Everything else passed

**The fix is genuinely present and complete in `933613d`.** `truncationDetail` is in the tree, the
guard is the first thing `volumeFindingForPair` does, `ukLocator` / `otherLocator` are correctly
hoisted above it, and `UA_VOLUME_PARITY_CLAUSE` is the pre-existing constant. Nothing is half-applied.

**The regression tests fail without the fix.** Reverting *only* `src/rules/ua.ts` to `origin/main` and
keeping the test file reproduces the described result exactly — `3 failed | 61 passed (64)`, all three
under `ua/state-language-version-lesser > a truncated text sample`, and in both directions:

| Test | Assertion | Direction |
| --- | --- | --- |
| `does not read parity from two samples that both hit the cap` | `expected [] to have a length of 1` | under-fires (missing finding) |
| `does not accuse a Ukrainian page whose own sample was the one truncated` | `expected [ {…} ] to deeply equal []` | false accusation |
| `does not accuse a Ukrainian page whose counterpart was the truncated one` | `expected [ {…} ] to deeply equal []` | false accusation |

The reverted run publishes the `92% less` `fail` with `Law 2704-VIII` / `Art. 27 §6` attached, which is
the defect verbatim. The two negative controls pass in both states, so the gate is not over-firing.

**No mode change crept back.** `git diff origin/main --summary` is empty, and the blob is `100644` in
both `origin/main` and `933613d`.

Worth recording for whoever hits this next: the artifact is **not** from running the audit CLI. It is
`pnpm install` — `packages/audit/package.json` declares `"bin": { "movar-audit": "./src/collect/cli.ts" }`,
and pnpm `chmod +x`es the bin target when it links the workspace. It reproduced in a clean worktree
during this review from `pnpm install` alone. Anyone who installs before committing will pick it up.

**The design is right, and it is the documented contract rather than a local invention.** `evidence.ts`
already specifies `cappedAt` this way:

> The ceiling that bit, set **only** when the sample was truncated. Its presence — not a comparison of
> the two counts — is how a rule asks "may I still compare these pages by volume?".

Keying on `examined > sampled` would be wrong on its own terms — that comparison can be true for
reasons other than a cap — and hard-coding `MAX_TEXT_NODE_SAMPLES` would put the collector's constant
inside a kernel that adjudicates bundles from collectors it has never seen.

On pre-v3: reading a missing report as untruncated does leave a truncated pre-v3 bundle compared the
old way. That is the only reading available, it matches the `schemaVersion` 3 note in `evidence.ts`
(the field degrades by *fallback*, not to `not-collected`), it is stated in the code, in `AGENTS.md`
and in the description, and it has a dedicated test. Reading an unknown as a truncation would silence
the rule on every stored bundle including small pages the cap never approached. Not silent, and not a
regression this PR introduces.

**`info` is the honest verdict.** Rule-level `not-collected` is not available and would be wrong: the
kernel decides that from capabilities before `run` is called, this rule may not hand-check for missing
data, and truncation is per-pair — a rule-level degrade would discard the pairs it *can* measure.
Within the pair the options are silence, `info`, or an accusation. Silence is the failure mode
`not-collected is never pass` exists to prevent, one layer down: a pair that was never measured would
be indistinguishable from one measured clean. An accusation is the bug being fixed, and `fail` counts
toward `brokenPromises`, so a collector ceiling could turn a build red. `info` is the only verdict left
that is true.

The summary holds up under the wording scrutiny this rule deserves. It names only the sides whose own
evidence carries `cappedAt`, quotes their `sampled of examined` counts, asserts no deficit, carries no
percentage (`expect(...).not.toMatch(/% less/)` guards that), and closes by explicitly disclaiming
assessment — *"this pair was not assessed against that requirement."* The `Art. 27 §6` citation is
mandatory under the grading law for pack findings, not a rhetorical choice. Every clause is true
whether one side or both were capped.

**No scope creep and no new invariant violation.** Exactly the six expected files. `ua.ts` stays pure —
no clock, randomness, `fetch`, filesystem, or DOM globals. `classified` is untouched. The new finding
is `declared`/`info`, so it cannot manufacture a build-breaking accusation.

**CI was green on `933613d`** — all six (`verify`, `build`, `audit-site`, `e2e-offline`, `ci-gate`,
`metrics-gate`) — but that run predates #470, so it no longer certifies the merged result.

### Also worth a follow-up (not blocking, already noted in the description)

`classifyDefaultLanguage` builds its `Denominator` from `nodes.length` — the filtered *sample* — rather
than going through `textNodeDenominator()`. On a truncated page `ua/state-language-not-default` quotes
the cap instead of the population, overstating the published share in the direction of the accusation,
on another statute-cited rule. Verified present; correctly out of scope here. There is no open issue
for it yet.

### To unblock

Rebase onto `d50d1c3`, regenerate the metrics snapshot under #470's `.astro` convention, force-push,
and re-run all six checks. The code needs no changes — I verified the merge is green.

PR #480, opened 2026-08-15
fix(audit): match robots.txt patterns per RFC 9309, not by prefix

### LINKED ISSUE #434: audit: robots.txt wildcards are matched literally and grouped user-agent lines are dropped

rejifald:
Closes #434

## The defect

`robots.txt` is honoured for declared-target expansion — the one place this collector does automated multi-page access — and three bugs in that path all failed **open**, so the collector fetched paths the site had asked it not to.

A and B live in `packages/audit/src/collect/probe.ts`; C is the call site in `packages/audit/src/collect/node.ts`. (The issue cites `probe.ts:541` / `:530`, written against `bd14285`; the code has moved, but the functions it names — `robotsAllows` and `parseRobots` — are the real ones and are still in that file.)

### A — `*` and `$` were matched as literal characters

`robotsAllows` asked `path.startsWith(pattern)`, so RFC 9309 §2.2.3's metacharacters never applied:

| `robots.txt` | path | before | after |
| --- | --- | --- | --- |
| `Disallow: /*/private` | `/uk/private` | fetched | withheld |
| `Disallow: /*.json$` | `/data.json` | fetched | withheld |
| `Disallow: /*.json$` | `/data.json.html` | fetched | fetched (the `$` anchors) |
| `Disallow: /*?` | `/search?q=1` | fetched | withheld |
| `Disallow: /private` + `Allow: /private/*/public` | `/private/uk/public` | withheld | fetched (longest pattern wins) |

### B — a run of `User-agent` lines depended on its order

`parseRobots` recomputed the flag per line (`inWildcardGroup = directive.value === '*'`), while §2.2.1 makes consecutive `User-agent` lines **one** group applying to all of them:

| `robots.txt` | before | after |
| --- | --- | --- |
| `User-agent: Googlebot` / `User-agent: *` / `Disallow: /private` | `disallow: ['/private']` | `disallow: ['/private']` |
| `User-agent: *` / `User-agent: Googlebot` / `Disallow: /private` | `disallow: []` — group dropped | `disallow: ['/private']` |

### C — the call site cut the query off before the matcher saw it

Fixing A is not enough to make `Disallow: /*?` work. §2.2.2 matches a rule against the path **and** the query, but `followDeclared` asked `robots.allows(new URL(target).pathname)` — so the `?` the pattern turns on was never in the string being matched, and the rule could not fire on any real declared target:

| declared target | matched against, before → after | before | after |
| --- | --- | --- | --- |
| `https://site/search?q=1` | `/search` → `/search?q=1` | fetched | withheld |
| `https://site/?lang=uk` | `/` → `/?lang=uk` | fetched | withheld |
| `https://site/uk/` | `/uk/` → `/uk/` | fetched | fetched (no query; the rule does not apply) |

Reachable, not theoretical: `AlternateLink.href` and `PickerOption.href` are stored verbatim and resolved with `new URL(href, base).toString()`, which preserves the query — and query-parameter language switching (`?lang=`, `?hl=`, `?locale=`) is exactly the shape a picker's declared targets take. Without C, `/*?` — the third of the three patterns #434 names, and the one it calls "very common" — would have stayed bit-for-bit as inert after this PR as before it.

## The fix

- `robotsAllows` now filters through `robotsPatternMatches`, which walks the literal runs between the `*`s with `startsWith` + `indexOf`. Longest-match-wins still reads the pattern's own length (metacharacters included), so `Allow`-over-`Disallow` precedence is unchanged, and an absent, empty or unparseable `robots.txt` still allows everything.
- `parseRobots` keeps an `inAgentRun` flag beside the group flag: a run of `User-agent` lines opens the group when *any* of its agents is `*`, and only a rule line ends the run — a blank line, a comment or a `Sitemap:` sits inside a group untouched, per the §2.2.1 ABNF.
- `followDeclared` matches against `robotsSubjectOf(target)` — `pathname` + `search` — instead of `pathname` alone, and `ResolvedRobots.allows` names its parameter `subject` so the contract is not documented as a bare path. The fragment stays out: it never reaches the server. Two properties the review pinned survive it — longest-match-wins reads the *pattern's* length, not the subject's, so precedence is untouched; and the `$` end sentinel is still sound, because `URL.search` percent-encodes NUL exactly as `URL.pathname` does (`new URL('https://h/p?a=\0b').search` is `?a=%00b`), so appending it still adds no place for a pattern to match.

### Avoiding the regex trap

A `robots.txt` is attacker-controlled text from a third-party origin, so translating a pattern to a `RegExp` is a denial-of-service gadget. Measured with the naive `*` → `.*` translation of `/a*a*…*b` (20 stars):

- naive `RegExp`, **41-character** path: **65 491 ms**
- the shipped scan, **50 000-character** path: **0.01 ms**

`indexOf` cannot backtrack, and leftmost-first is optimal when `*` is the only metacharacter — an earlier match never rules out a later run. A `$` is handled by appending an end sentinel (NUL, which neither a URL path nor a URL query can contain) to pattern and subject alike, so the anchored case is the same prefix scan rather than a second code path. The implementation was differentially fuzzed against a port of Google's reference `robots.cc` matcher: 500 000 random pattern/path pairs over `/ a b . ? * $`, zero disagreements.

## The regression tests

Seven new cases in `packages/audit/src/collect/probe.test.ts`, all of which fail on `main` and pass here:

```
× expands a * inside a Disallow pattern
× anchors a Disallow pattern ending in $ to the end of the path
× anchors the start of a $-terminated pattern too
× honours the query-string idiom Disallow: /*?
× keeps longest-pattern-wins when the longer pattern holds a wildcard
× reads a group whose consecutive user-agent lines start with the wildcard
× reads consecutive user-agent lines the same way in either order
```

The first four are the issue's own reproductions (`/uk/private`, `/data.json`, plus the `/*?` idiom it calls out); the fifth pins the precedence rule that must survive the new matcher; the last two are part B, one asserting the parse and one asserting the two orderings agree. Two further cases guard the fix's own failure modes and pass either way: `matches every other pattern character literally` (`/a.b(c)+d` must not behave like a regex) and `starts a new group at the first user-agent line after a rule`.

One new case in `packages/audit/src/collect/node.test.ts` covers C, and covers it **end to end at `followDeclared`** rather than at the matcher — which is the whole point of it, since the matcher already read `/*?` correctly and the call site threw the query away before it got there:

```
× withholds a query-carrying declared target under Disallow: /*?
  AssertionError: expected [ 'https://example.com/', …(2) ] to not include 'https://example.com/?lang=uk'
```

It collects a page declaring two alternates — `/?lang=uk` and `/de/` — under `Disallow: /*?`, then asserts the query-carrying target appears in neither the probe list nor the page list, while the query-free sibling is still followed (so this is not a blanket withholding). Asserting on the **probes** is the load-bearing half: `robots.txt` refuses the *request*, and a target merely dropped after fetching it would already have been the access the site declined.

That the assertion has to sit here is demonstrable: reverting only the call-site expression to `new URL(target).pathname` fails exactly this one test and leaves all 38 matcher cases in `probe.test.ts` green.

`packages/audit/AGENTS.md` gains a gotcha recording all three parts and the no-`RegExp` rule, since the next work on this code (#429, per-origin `robots.txt`) touches both `parseRobots` and `followDeclared`.

## Verification

- `pnpm --filter @movar/audit test` — 25 files, 823 tests, green
- `pnpm validate` (including `audit:site`) — green; no ruleset or site change, so no suppression was added and none went stale
- `pnpm metrics:audit` (`fallow audit --base origin/main`, the pre-push complexity gate) — no issues in the 7 changed files; `fallow --score` reports 81.9 / **B**, unchanged, with `functions_above_threshold: 0`
- `pnpm test:coverage && pnpm gen:readme --refresh` — `coverage.lines` 96.6 and `coverage.branches` 90.5 both hold; `loc` 60811 → 60826
- `git diff origin/main --summary` is empty — no mode changes, no adds or deletes, exactly the 7 expected files


rejifald:
## Review: blocked on one finding

The fix itself is good work and most of it verified cleanly. Blocking on scope, not on correctness.

### Blocking — `Closes #434` is false, and the follow-up it claims does not exist

Issue #434 names three patterns. Two are genuinely fixed end to end. The third is not, and the PR's own "Not in scope" note concedes it:

> `followDeclared` … asks `robots.allows(new URL(target).pathname)`, which drops the query string — so a `Disallow: /*?` rule is honoured by the matcher but still cannot fire on a real declared target.

Running the exact call-site expression from `packages/audit/src/collect/node.ts:193` against `parseRobots('User-agent: *\nDisallow: /*?')`:

```
target https://h/search?q=1       pathname=/search   allows=true   <-- FETCHED
target https://h/?lang=uk         pathname=/         allows=true   <-- FETCHED
target https://h/page?lng=en&x=2  pathname=/page     allows=true   <-- FETCHED

matcher alone, '/search?q=1'                        allows=false   (correct)

/*/private vs https://h/uk/private                  allows=false   (fixed)
/*.json$   vs https://h/data.json                   allows=false   (fixed)
```

So `Disallow: /*?` is bit-for-bit as inert after this PR as before it. Zero behavioural change for the one pattern #434 singles out as "very common". The issue's stated harm — "the collector fetches paths the site asked it not to" — is still live for it.

This is reachable, not theoretical. `AlternateLink.href` and `PickerOption.href` are stored verbatim, `safeUrl` is `new URL(href, base).toString()` which preserves the query, and only `node.ts:193` strips it:

```
href=/?lang=uk       -> declared target https://site.com/?lang=uk   -> pathname passed to robots: /
href=?hl=uk          -> declared target https://site.com/en/?hl=uk  -> pathname passed to robots: /en/
```

Query-parameter language switching (`?lang=`, `?hl=`, `?locale=`) is exactly the shape a language picker's declared targets take.

**What decides it:** the PR says this was "reported for its own issue rather than widened into this fix." It was not. There is no such issue — searching all open and closed issues for robots turns up only #429 (cross-origin robots.txt, a different defect) and #434 itself. Merging auto-closes a `bug`+`security` issue with a third of its named surface still broken and nothing anywhere tracking it.

Either fix is fine by me:

1. Pass `${u.pathname}${u.search}` at `node.ts:193` (RFC 9309 §2.2.2 matches path *and* query) with a `followDeclared`-level regression test — the matcher already handles it, and longest-match-wins reads pattern length so precedence is untouched; or
2. Keep the scope split, drop `Closes #434` to a plain reference, and file the follow-up issue for real so #434 stays open until its third pattern is dead.

### Everything else verified and passing

- **Fix is real**, not a tests-only diff: `probe.ts` +60/-6 of production code.
- **Pre-fix failure split is exactly as claimed** — reverting only `probe.ts` (worktree only, index untouched) fails exactly the 7 named tests; the 2 named guards pass both ways.
- **Mutation kill matrix** (38 tests, baseline green):

  | mutation | tests killed |
  | --- | --- |
  | naive `*`→`.*` regex | `/*?` idiom **+** literal guard |
  | correct fully-escaped regex (control) | none |
  | `(`, `)`, `+` regex-live, rest escaped | **literal guard only** |
  | delete `inAgentRun = false` | **grouping guard only** |
  | restore original per-line group flag | the 2 part-B regression tests only |
  | `$` treated literally | the 2 `$` tests only |
  | delete `if (kind === 'skip') continue` | 2 pre-existing tests |

  Both guards earn their place: each is uniquely killed by a mutation nothing else catches, and neither is redundant with the regression tests. (The literal guard is not *solely* killed by the naive-regex mutation — the `/*?` test independently catches that too, since `?` is a regex metacharacter and `^/.*?` matches empty. That is a strengthening.) The new `skip` early-`continue` is load-bearing: without it `directive.value` is `undefined` and gets pushed into a bucket.

- **No `RegExp` claim holds.** Nothing on this path builds a `RegExp` from robots.txt content. The only regex in the region is the constant line splitter `text.split(/\r?\n/u)` — fixed pattern, linear, no backtracking gadget. One production caller, `node.ts:193`.
- **Backtracking numbers reproduced independently**: naive `RegExp` on `/a*a*…*b` (20 stars), 41-char path — **65 816 ms**; shipped scan — **0.005 ms** at 41 chars, **0.008 ms** at 50 000 chars.
- **NUL sentinel holds.** `new URL('https://h/a\0b').pathname` is `/a%00b`; the only caller passes `URL.pathname` and nothing on this path decodes it (`decodeURIComponent` appears only in `url-language.ts`, unrelated). And the crafted case fails *closed*: `/secret\0extra` against `Disallow: /secret$` is withheld, so a sentinel can only over-match. `$` anchoring is not defeatable.
- **RFC 9309 spot checks: 17/17**, including `*` at end (`/fish*`), `$` mid-pattern staying literal (`/a$b` matches `/a$b`, not `/ab`), empty pattern, `*` alone, and `/*`.
- **Grouping is RFC §2.2.1-correct** on everything I threw at it: both two-agent orderings agree; three consecutive agents with the wildcard in each of the three slots all yield the group, and a three-agent run with no wildcard yields nothing; a `User-agent` after a rule starts a new group in both orderings; blank lines, comments and `Sitemap:` neither end an agent run nor end a group; CRLF and mixed-case fields fine.
- **No scope creep.** `git diff origin/main <head> --summary` is empty — no mode changes, no adds or deletes, exactly the 5 expected files.
- **Metrics snapshot is a genuine recompute.** Ran `pnpm test:coverage` then `gen-readme-metrics.mts --refresh` from scratch: **zero diff** against the committed snapshot. `coverage.branches` 90.5 and `loc` 60811 both reproduce.
- **Invariants intact.** Pure string work in `collect/`, no new I/O, no verdict-derivation change. The added restrictiveness is honest in its direction: a withheld target means fewer `declared-target` pages, so a traversal rule may read `not-collected` — which is never `pass`, and is the correct answer for evidence the site forbade collecting. Strictly better than the pre-fix behaviour of producing a verdict from pages fetched against robots.txt.

  Minor, pre-existing, not for this PR: a withheld target is silently `continue`d with no `ProbeEvidence` pushed, so the bundle records `robots.posture: 'honoured'` but not *which* targets were withheld — a reader cannot tell that a given rule went `not-collected` because of robots. This PR widens how often that is hit without introducing it.

- **CI**: all six green on `abee829`, and fresh — main's tip `777d61f` (00:33:03Z) predates the run start (00:52:15Z), and the PR head is a direct child of current main. Not stale.

Issue #436, opened 2026-08-14
audit: a declared target that dedupes onto a collected page never grants traversal


rejifald:
`digestInto` returns the id of an already-collected page without upgrading its `reach`, so a declared target that resolves to a page the matrix already fetched leaves `traversal` underived. Every traversal rule then reports `not-collected` even though `--follow` ran and followed every declared target successfully.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `bd14285`.

## What happens

`packages/audit/src/collect/node.ts:203`:

```ts
const existing = seenDocuments.get(key);
if (existing !== undefined) return existing;
```

The page keeps `reach: 'requested'`. `packages/audit/src/capability.ts:hasTraversal` looks for `page.reach === 'declared-target'`, finds none, and withholds `traversal`.

## Reproduction

A single page whose only declared alternate is the ubiquitous self-referential `<link rel="alternate" hreflang="en" href="https://example.com/">`, collected with `followDeclaredTargets: true`:

```
fetched: [ 'https://example.com/', 'https://example.com/' ]
pages:   [ 'https://example.com/ [requested]' ]
capabilities: [ 'static', 'http' ]
core/hreflang-target-unresolvable => not-collected [ 'traversal' ]
```

The second fetch is also a wasted request against the budget.

## Why it matters

The honest answer is `pass` — the collector followed every declared target and they all resolved. Instead the report says the collector lacked the capability, which is the same class of dishonesty `not-collected is never pass` exists to prevent, pointing the other way.

Affects `core/hreflang-target-unresolvable`, `core/hreflang-target-wrong-language`, `core/picker-target-unresolvable`, `core/switch-no-effect`, `core/switch-loses-path`, `core/switch-bounces`.

## Fix sketch

Track reach separately from the dedupe: when a declared-target probe resolves onto an existing page, upgrade that page's `reach` to `'declared-target'` (reach is a property of how the page was *reached*, and it was reached both ways). Skipping the redundant fetch when the target is already collected would also return the spent request to the budget.

PR #481, opened 2026-08-15
fix(audit): resolve robots.txt per origin, not once for the typed URL

### LINKED ISSUE #429: audit: the collector applies the first origin's robots.txt to cross-origin declared targets, and never reads theirs

rejifald:
Closes #429

## The defect

`resolveRobots` in `packages/audit/src/collect/node.ts` fetched `robots.txt` from the origin of the URL the operator typed, and `followDeclared` applied those rules to **every** declared target — including hreflang alternates on another origin, which is the standard cross-domain locale pattern (`brand.de` / `brand.pl` / `brand.ua`). A robots.txt binds the origin that serves it and no other, so one document was being read over all of them, wrong in both directions at once.

**A — the wrong site's rules withheld a legitimate target, and the collector then accused the site of it.** With `example.com/robots.txt` = `Disallow: /de`, the declared alternate `https://example.de/de/` was never fetched, and the gap that left in the page set published `core/hreflang-target-unresolvable` as a `fail`:

> The hreflang="de" target "https://example.de/de/" is absent from the collected page set, so the declared alternate cannot be reached.

example.de serves that page perfectly well and its own robots.txt permits it. That is a false accusation about a named company, which `docs/movar-audit.md` identifies as the one failure mode that ends the product.

**B — the other origin's robots.txt was never requested at all**, so `resolveRobots`' own doc claim that robots.txt is "honoured for declared-target expansion, which is automated multi-page access" was true of the first origin and no other.

## Before / after

| | before | after |
| --- | --- | --- |
| `example.de/de/` under `example.com`'s `Disallow: /de` | withheld; `core/hreflang-target-unresolvable` **fails** | fetched; rule **passes** |
| `example.de/robots.txt` | never requested | requested once, before the first target on that origin |
| `example.de/de/` under `example.de`'s own `Disallow: /de` | fetched — the site's refusal was never read | withheld |
| N targets on one origin | — | one robots.txt, cached; failures cached too |
| `--ignore-robots` | no robots request | unchanged: no robots request |

## The fix

`createRobotsGate` keys `new URL(target).origin` into a per-origin cache and evaluates each target against its **own** origin's rules. The RFC 9309 matcher and `robotsSubjectOf` from #480 are reused unchanged — there is still exactly one matcher, and it still receives path **and** query.

- **Caching.** One `Map` per run, keyed on origin, populated on the first target of each origin. Refusals *and* failures are stored, so N targets on one origin cost one request and an unreachable robots.txt is not re-asked per target.
- **Unfetchable stays permissive.** `EMPTY_ROBOTS` for a probe that is not `ok`, has no body, or throws — the same semantic the previous code had (`allows: () => true`), verified before it was carried over.
- **Lazy.** A run with no declared targets, or one whose budget the matrix already spent, fetches no robots.txt at all, where `main` always bought one.
- **No phantom origins.** A target with no fetchable origin — a picker's `javascript:` href resolves to an opaque one — is neither asked about nor charged for; it is probed exactly as it was before.

## The budget policy — rewritten after review

The per-origin fetch is a real request against the same hard budget: the honest price of the posture the ADR describes. The first revision paid for it with a reserve — the gate withheld a target while `remaining() < 2`, so a permission slip could never be the last request. Review found that unsound in **both** directions, and both are fixed here.

### 1. A reserve cannot be sized, so there is none

`prober.probe()` spends one request **per redirect hop** — `spend()` sits inside `walk()`'s loop, not outside it — so `remaining() >= 2` never bought "robots.txt plus target". It bought two hops. A `robots.txt` that 301s (`http`→`https`, `www`→apex, CDN normalisation — ordinary) consumed the reserve, `fetchRobots` swallowed nothing because the redirect succeeded, the gate returned `true`, and the probe immediately behind it threw `RequestBudgetExhaustedError` out of `collectNetwork` — on a run that completes on `ffce2c0`. A strict regression, not a pre-existing hazard: `main` was safe by construction, because its robots fetch preceded the loop whose top re-checks `remaining() === 0`; moving the fetch *inside* the iteration, after that check, removed the guarantee.

Restored by re-reading the ceiling rather than pre-computing it. `followDeclared` now checks `prober.remaining() === 0` **after** the gate returns and before the probe — the only check that can be right when the hop count is not knowable before the fetch. Any reserve is either too small (the probe behind it throws) or too large (a target the budget could pay for is withheld). `fetchRobots`' existing `catch` reads a chain that reaches the ceiling mid-redirect as "no rules published", and the re-check ends the expansion before anything is fetched on the strength of rules nobody read.

The gate therefore has no budget branch at all, and its `false` now means exactly one thing: **the site said no.**

### 2. A scarce budget buys the pages the audit does not already hold

The reserve did refuse targets the budget could have paid for, and a withheld target is not a silence — `unresolvableDraft` publishes it as `fail … "cannot be reached"` about a site that permits and serves the page. That is the same false accusation #429 exists to kill, re-created on the budget path.

`budgetOrderOf` puts declared targets the page set does not already hold ahead of ones it does. A target already collected can only add a `reach` upgrade; a target absent from the set is the one the report will otherwise call unreachable. Spending in markup order let the ubiquitous self-referential `<link rel="alternate" hreflang="en" href="/">` crowd out the cross-origin alternate behind it — with `budget: 4` on the review's case, `ffce2c0` collects `https://example.de/de/` and publishes `pass`, the reserve left a request unspent and published the `fail`, and this branch now collects it and publishes `pass` with the ceiling reached. Stable within each half, so a run stays reproducible.

Consequences, stated rather than hidden:

- At most one robots.txt per run can end up buying nothing — the one the last request pays for. `main` did that unconditionally, buying a robots.txt before the loop whether or not the budget could use it.
- A redirecting **target** at `remaining() === 1` can still throw. That is pre-existing on `main` and not this PR's to fix.
- The gate still cannot tell the report *why* a target is missing. Withheld-by-the-site and never-reached-at-all both read "absent from the collected page set", which is the misattribution the review carved out as a separate defect: it predates this PR (it reproduces on `main` with a same-origin alternate under `Disallow:`) and closing it needs a withheld-target signal in the evidence schema. Pinned in the suite here rather than left to be discovered in a report about a real company.

## How the tests prove it

All in `packages/audit/src/collect/node.test.ts`, at `collectNetwork` rather than at the matcher — a matcher-level assertion would have missed this exactly as it would have missed #434. **Every budget case adjudicates `CORE_RULESET` and asserts the published verdict**, not just the fetch log: the two budget tests in the first revision asserted request lists only, which is precisely why a verdict flipping from `pass` to `fail` about a named company was invisible.

Per-origin resolution, failing on `ffce2c0` before the fix:

- **`follows a cross-origin target its own origin permits`** — direction A end to end: asserts the page is collected *and* that `core/hreflang-target-unresolvable` reads `pass`. Pre-fix: `expected [ 'https://example.com/' ] to include 'https://example.de/de/'`, and the rule published the `fail` quoted above.
- **`requests each origin's own robots.txt`** — direction B, on the fetch log. Pre-fix: `expected [ 'https://example.com/', …(3) ] to include 'https://example.de/robots.txt'`.
- **`withholds a cross-origin target its own origin disallows`** — the flip side, so the fix is per-origin evaluation and not "skip robots for other origins". Pre-fix the target was fetched despite example.de refusing it.
- **`fetches an origin robots.txt once, even when it is unreachable`** — two targets on one origin, whose robots.txt cannot be reached: exactly one attempt, and both targets still followed. Pre-fix: zero attempts.

The budget policy, failing on this branch's previous head (`81257ec`) before this revision:

- **`survives a robots.txt that redirects into the last of the budget`** — `budget: 3`, `example.de/robots.txt` 301s once. Pre-fix: `RequestBudgetExhaustedError: request budget of 3 exhausted …` thrown out of `collectNetwork` (`followDeclared` → `probe` → `spend`). Now the run completes: `[HOME, CROSS_ROBOTS, CROSS_ROBOTS_MOVED]` fetched, `https://example.com/` collected, and the traversal rule reads `not-collected` — the audit says what it did not fetch instead of accusing anyone of it.
- **`buys the cross-origin page before re-fetching one it already holds`** — `budget: 4`, allow-all everywhere. Pre-fix: `expected [ 'https://example.com/' ] to include 'https://example.de/de/'`, and the rule published the `fail`. Now the page is collected, the rule reads `pass`, and `prober.remaining()` is `0` — nothing is withheld while a request goes unspent.
- **`spends its last request asking permission rather than leaving it unspent`** — `budget: 2` (was `holds a request back rather than spending the last one on robots.txt`, whose premise the reserve owned). The run ends cleanly at its ceiling, and the verdict is asserted: `not-collected`, never a `fail` naming the site.
- **`spends the budget on a robots.txt and the target it authorizes`** — `budget: 3`, extended to adjudicate the ruleset. The second target is withheld for budget and the report says `fail` about `example.de/at/`: the misattribution defect above, pinned so its cost is visible in the suite.

Pinning the rest of the contract:

- **`asks no robots.txt for a target with no fetchable origin`** — a `javascript:` picker entry costs no request.
- **`records an explicit ignore-robots posture`** — extended to assert no `robots.txt` is requested at all under `--ignore-robots`.

`packages/audit/AGENTS.md` gains the per-origin invariant beside the RFC 9309 matcher bullet, plus a second one for the budget: re-read never reserved, pages-first ordering, and "test every budget path against the ruleset, never the fetch log alone".

## Verification

`pnpm --filter @movar/audit test` — **832 pass** (830 on the previous head; the description's earlier "829" was wrong, as review noted). `pnpm validate` green with `audit:site` included. `pnpm test:coverage` + `pnpm gen:readme --refresh`: coverage holds at **96.7 lines / 90.6 branches**, snapshot committed (LOC only moved). `pnpm metrics:audit` clean — no suppression added, nothing pinned, `git diff origin/main --summary` empty.


rejifald:
## Review: blocking on the budget reserve

The per-origin resolution itself is right and I verified it end to end. Both directions of #429 reproduce on `ffce2c0` with exactly the messages claimed, and both are genuinely cured here:

- **A** — `follows a cross-origin target its own origin permits`: pre-fix `expected [ 'https://example.com/' ] to include 'https://example.de/de/'`; and evaluating `CORE_RULESET` on that evidence, `core/hreflang-target-unresolvable` is **`fail`** before and **`pass`** after, with `https://example.de/de/` actually in `evidence.pages`. Not a different wrong answer — the page is collected.
- **B** — `requests each origin's own robots.txt`: pre-fix `expected [ 'https://example.com/', …(3) ] to include 'https://example.de/robots.txt'`.
- Plus `withholds a cross-origin target its own origin disallows` and `fetches an origin robots.txt once, even when it is unreachable`, both failing pre-fix as claimed. Reverted with `git show origin/main:… > …`, unstaged.

A genuinely unreachable target still reads `fail` with the right summary, caching holds (a **refusing** `robots.txt` is asked once for two targets on that origin, and both are withheld), resolution is genuinely lazy (a run with no declared targets now fetches **no** `robots.txt` at all, where `main` always bought one), `--ignore-robots` short-circuits on the first line of the gate before `robotsOriginOf`, an opaque-origin href costs nothing, `EMPTY_ROBOTS` preserves `main`'s permissive semantics exactly, scope is clean (`git diff origin/main --summary` empty), and the snapshot reproduces from a fresh `pnpm test:coverage && pnpm gen:readme --refresh` with no drift. All six checks are green on the current head and `main` has not moved.

The block is the `remaining() < 2` reserve. Two findings, one of them a crash.

### 1. Blocking — `RequestBudgetExhaustedError` escapes `collectNetwork`; the reserve of 2 is unsound

`prober.probe()` spends one request **per redirect hop** — `once()` calls `spend()` inside `walk()`'s loop — not one per probe. So `remaining() >= 2` does not buy "robots.txt plus the target". A `robots.txt` that redirects (`http`→`https`, `www`→apex, CDN normalisation — very common) consumes the reserve, `fetchRobots`' `catch` swallows nothing because the redirect succeeded, the gate returns `true`, and the probe immediately behind it walks into exactly the error the comment says it prevents.

```ts
// budget: 3, headers: [null]
// https://example.com/  → <link rel="alternate" hreflang="de" href="https://example.de/de/">
// https://example.de/robots.txt  → 301 → https://example.de/robots2.txt → 200 "User-agent: *\nDisallow:"
// https://example.de/de/         → 200
```

- on `ffce2c0`: completes; fetched `["https://example.com/", "https://example.com/robots.txt", "https://example.de/de/"]`; pages `["https://example.com/", "https://example.de/de/"]`
- on this branch: **throws** `RequestBudgetExhaustedError: request budget of 3 exhausted …` out of `collectNetwork`, after `["https://example.com/", "https://example.de/robots.txt", "https://example.de/robots2.txt"]`

This is a strict regression, not a pre-existing hazard. `main` was safe by construction: the robots fetch happened *before* the loop, and the loop re-checks `remaining() === 0` at the top of every iteration, so a robots fetch that ate the budget just ended the expansion. Moving the fetch *inside* the iteration, after that check, removes the guarantee. The smallest honest fix is to re-check after the gate rather than trust an arithmetic reserve:

```ts
if (!(await allows(target))) continue;
if (prober.remaining() === 0) break; // the gate may have spent more than one hop
```

(A redirecting *target* at `remaining() === 1` can still throw — that one is pre-existing on `main` and not this PR's to fix.)

### 2. Blocking — the reserve withholds a target that `main` fetched, and the report accuses the site for it

The reserve does refuse targets that could have been fetched: at `remaining() === 1` on an origin not yet in the cache, the last request goes unspent and the target is dropped. Where some *other* declared target already granted `traversal`, that drop is published as a `fail` naming the site:

```ts
// budget: 4, headers: [null]
// https://example.com/ declares hreflang="en" → https://example.com/  and  hreflang="de" → https://example.de/de/
// every robots.txt allow-all; https://example.de/de/ serves 200
```

- on `ffce2c0`: pages `["https://example.com/", "https://example.de/de/"]`, `core/hreflang-target-unresolvable` = **`pass`**
- on this branch: pages `["https://example.com/"]`, one request left unspent, and

  > **fail:** The hreflang="de" target "https://example.de/de/" is absent from the collected page set, so the declared alternate cannot be reached.

`example.de` serves that page and its own `robots.txt` permits it. That is the same false accusation about a named company #429 is about, re-created on the budget path — an input `main` reported correctly and this branch reports wrongly. The two budget tests here (`holds a request back …`, `spends the budget on …`) assert fetch logs only; neither evaluates the ruleset on the resulting evidence, which is why the flip is invisible.

I don't think the *gate* is wrong — at `remaining() === 1` on an unresolved origin there is no honest way to both ask permission and fetch, and withholding beats crawling a page you have no permission slip for. The wrong part is that `Evidence` carries no record that the collector withheld the target, so `unresolvableDraft` can only say "absent from the collected page set … cannot be reached". Please at minimum add a case that pins the published verdict for a budget-withheld target, so the cost is visible in the suite rather than discovered in a report about a real site.

### `Closes #429` — TRUE, and the adjacent defect is genuinely separate

Both symptoms #429 states are dead: the cross-origin permitted target is collected and reads `pass`, and `example.de/robots.txt` is requested. Ruling that the misattribution defect is separate, and I verified it rather than taking it on trust:

- On **`ffce2c0`**, with `example.com/robots.txt` = `Disallow: /uk` and a same-origin `hreflang="uk"` alternate, `core/hreflang-target-unresolvable` already publishes `fail: The hreflang="uk" target "https://example.com/uk/" … cannot be reached.` The "collector's own restraint attributed to the site" defect predates this PR; it is not introduced here.
- `unresolvableDraft` in `rules/inventory-hreflang.ts` distinguishes only 404 from "absent from the collected page set" — closing it needs a withheld-target signal in the evidence schema, which is a different change from per-origin resolution.
- The "no existing issue" claim checks out: I searched all 126 open and closed issues by title and by body (`robots`, `cannot be reached`, `withheld`, `traversal budget`). #430 is relative-hreflang resolution, #436 is dedupe/traversal, #434 is the matcher. Nothing covers this. It deserves its own issue.

One consequence worth stating in that issue: honouring an origin's own refusal is correct, but it moves cases from `pass` to `fail` that `main` got "right" only by ignoring the site's wishes — e.g. a cross-origin target whose own `robots.txt` says `Disallow: /de` now reads `fail … cannot be reached`. The collector behaviour is the improvement; the report wording is the debt.

### Not blocking

- The gate's `false` is ambiguous between "the site refused" and "the budget refused", which is what makes both of the above hard to see from the call site. A discriminated result would make the withheld-vs-refused distinction available to the evidence schema later.
- A robots fetch spent on an origin whose target then turns out to be disallowed still costs a request that a *different* origin's target could have used. Inherent to the posture, worth knowing.
- 830 tests pass locally, not the 829 in the description.

Everything else here I'd merge as is.

PR #484, opened 2026-08-15
fix(audit): take the two hybrid seams' denominators from what was examined

### LINKED ISSUE #483: audit: two published denominators still quote the sampling cap as the page, after #438 fixed the other five
#438 put the sampling report on the wire and migrated the five `textNodeDenominator` call sites. Two published denominators were missed, because they build their own `Denominator` locally instead of calling it — so on a capped page both still quote the 1500-node cap as if it were the page, understating `M` and inflating the share the finding publishes.

Found by an automated bug-hunt sweep of `packages/audit`; verified against `f8a7609`.

## What happens

Neither of these routes through `textNodeDenominator`, so neither can see `textSampling.examined`:

- `packages/audit/src/served-language.ts:79` — `classifiedPageLanguage()` returns `denominator: { examined: samples.length, matched }`, where `samples = page.document.textNodes`. Reaches the report through `core/switch-no-effect` (`packages/audit/src/rules/switch.ts:229`) and, via `mergedDenominator`, through family C's hybrid `core/serving-*` findings.
- `packages/audit/src/rules/ua.ts:511` — `classifyDefaultLanguage()` returns `examined: nodes.length`, where `nodes` is `page.document.textNodes` filtered to non-blank. Reaches the report through `ua/state-language-not-default` (`packages/audit/src/rules/ua.ts:555`) — a statute-cited (Law 2704-VIII Art. 27 §6) `fail` published about a named company. The higher-stakes of the two.

Both summaries also still say *"sampled text nodes"*, the wording #438 dropped from the other five precisely because it stops being true of the wider count.

## Reproduction

A page whose walker examined 4000 eligible text nodes, of which the cap let 2 through, both classifying `ru`:

```
ua/state-language-not-default
  denominator: {"examined":2,"matched":2}   ← should be {"examined":4000,"matched":2}
  summary:     "…classifies as ru (2 of 2 sampled text nodes) rather than Ukrainian."

core/switch-no-effect  (examined 900, sampled 2)
  denominator: {"examined":2,"matched":2}   ← should be {"examined":900,"matched":2}
  summary:     "…and 2 of 2 sampled text nodes classify as ru — following the switch does not change the language."
```

## Why it matters

`finding.ts` states the standard: *"3 of 340 text nodes" is a finding; "3 nodes" is a smear.* `2 of 2` is not a smear, it is worse — it reads as a total, unanimous verdict on the page. The error runs in the direction of the accusation, and `ua/state-language-not-default` is where it lands hardest: the finding carries `UA_CITATION` and names a company.

`packages/audit/AGENTS.md` already records the rule this breaks — *"`textNodes` is a sample; `textSampling.examined` is the population… Take the denominator from `textSampling.examined` — `textNodeDenominator()` does"* — so this is the same defect #438 describes, in the two places that never called the function.

## Fix sketch

Route both through `textNodeDenominator(page, matched)` and take the summary numbers from the denominator the finding carries, as #438 did for the other five; drop "sampled" from both sentences. `classifyDefaultLanguage`'s blank filter narrows only what is handed to the classifier — the numerator over the sample against the widest honest population is the same pairing `classifySamples` already yields, and a conformant collector's `examined` excludes blank nodes anyway.

rejifald:
#475 put the sampling report on the wire and migrated the five `textNodeDenominator` call sites. Two published denominators were missed, because they assemble a `Denominator` literal instead of calling it — so both went on measuring the numerator against `page.document.textNodes.length`, the 1500-node sampling cap, as if it were the page. That understates `M` and thereby **inflates** the share the finding publishes, in the direction of the accusation: exactly the defect #438 describes, in the two places the migration never reached.

- `served-language.ts`'s `classifiedPageLanguage` — reaches the report through `core/switch-no-effect`, and through family C's hybrid `core/serving-*` findings via `mergedDenominator`.
- `ua.ts`'s `classifyDefaultLanguage` — reaches the report through `ua/state-language-not-default`, a `fail` that carries `UA_CITATION` (Law 2704-VIII Art. 27 §6) and names a company. The higher-stakes of the two: on a page whose walker saw four thousand eligible nodes it published *"2 of 2 sampled text nodes"*, which does not read as a smear but as a unanimous verdict on the whole page.

## What changed

Both now route through `textNodeDenominator(page, matched)`, and both summaries take their numbers from the denominator the finding carries, so the sentence and the measurement cannot drift apart. Both drop "sampled", which stops being true of the wider count — `finding.ts`' canonical phrasing is *"3 of 340 text nodes"*, and `viaNote`'s "served language read from sampled text" still says the text was sampled, because that clause is about the reading rather than the count.

`classifyDefaultLanguage`'s blank filter stays, and stays on the numerator only. It narrows what is handed to the classifier exactly as `classifySamples` narrows for the core families, and a conformant collector's `examined` already excludes blank nodes (`digest-dom.ts` skips anything under `MIN_SAMPLED_CHARS` before it increments the counter), so the two agree on any uncapped page. Counting the numerator over the sample against the widest honest population is the pairing every other classified denominator already uses, and it errs low — the safe direction for an accusation.

A bundle written before `schemaVersion` 3 carries no counts and falls back to the sample length, which is exactly what both quoted before.

## Tests

One regression test per rule, driven through `evaluate()` and asserting both the denominator and the summary text, following `content-language.test.ts`'s `measures a truncated sample against what was examined, not what survived`. Without the source change, four assertions fail:

```
ua/state-language-not-default    expected { examined: 2, matched: 2 } to deeply equal { examined: 4000, matched: 2 }
core/switch-no-effect            expected { examined: 2, matched: 2 } to deeply equal { examined:  900, matched: 2 }
```

plus the two updated summary assertions. `pnpm --filter @movar/audit test` → 834 passed; `pnpm validate` → green.

## A note on the `loc` snapshot

`scripts/readme-metrics.snapshot.json` moves `60917 → 60937`, which is exactly this PR's +20 source lines. Refreshing after `astro sync` would have written `61147` instead: the extra +210 is `apps/marketing/.astro/`'s generated, gitignored types, which `scanSourceStats` counts as source. That is #472, and the committed figure at every recent commit is the clean-tree value, so this one keeps that basis rather than baking the artefact into a published badge.

Closes #483


rejifald:
Superseded by #503 (`12f2a83`), which landed the same fix plus the `ua.ts` half.

Verified against merged main before closing: the buggy pattern `denominator: { examined: samples.length, matched }` is gone from both `served-language.ts` and `rules/ua.ts`, and this branch's versions of those lines are byte-identical to what main now has. `ua.ts` differed only in `const { denominator } = classified` vs `classified.denominator`.

The one thing not carried across is this PR's `packages/audit/AGENTS.md` edit, and that was deliberate. It stated an absolute rule — *"never assemble a `Denominator` literal, however local it looks"* — which is falsified in-tree by `rules/content-language.ts:349`, where `headDraft` correctly assembles one whose `examined` counts head fields rather than text nodes; `textNodeDenominator` would be wrong there. #496 independently arrived at the same over-broad wording, so the correctly-scoped version (a *body-text* denominator comes from `textNodeDenominator()`, never a literal) is worth writing down properly rather than lifting either draft.

Closed as part of an automated review sweep; reopen if you disagree.

PR #486, opened 2026-08-15
fix(audit): cite a text passage by itself, not the element around it

### LINKED ISSUE #447: audit: sibling text nodes collapse onto one nodePath, so findings cannot cite distinct passages
A text sample's `nodePath` is its **parent element's** path, so every text node under one element shares a single path. Findings about different passages then carry byte-identical `subject` and `evidence`, and a reader cannot tell which passage is meant.

## What happens

`packages/audit/src/collect/digest.ts:399`:

```ts
nodePath: parent === null ? '' : nodePathOf(parent, cache),
```

## Reproduction

`<p>A<br>B<br>C</p>` — three sampled text nodes:

```
samples: [{"p":"html > body > p","t":"Перше речення тут "},
          {"p":"html > body > p","t":"Друге речення теж "},
          {"p":"html > body > p","t":"Третє речення."}]
distinct nodePaths: 1 for 3 samples
```

## Why it matters

`evidence.ts` describes `NodePath` as "a stable pointer to an element inside a collected page (a CSS-ish path)", notes every collector must produce the same string for the same element, and `EvidenceRef { kind: 'node' }` is how a published finding tells a site owner *which passage* it is about.

`core/lang-part-unmarked` can emit several findings whose `subject` and `evidence` are indistinguishable — the report cites the same location three times for three different passages. `core/content-language-mixed`'s `citedNodes` has the same property.

## Fix sketch

Either extend `NodePath` with a text-node ordinal (e.g. `html > body > p :: text(2)`) — `evidence.ts` already anticipates this: "the most likely field to gain structure … once a second collector lands" — or add a separate index field to `TextNodeSample`. Any change here is a wire-format change and must be additive, since node paths are quoted in published findings.

rejifald:
Closes #447

## The defect

A `TextNodeSample`'s `nodePath` was its **parent element's** path, so every
passage under one element shared a single pointer. That is not an exotic shape:
a paragraph with an inline link in it is two sibling text nodes under one `<p>`,
which is ordinary prose.

**The issue understates it.** It reports "byte-identical `subject` and
`evidence`". In fact the whole published finding is a duplicate. On
`<p>RU passage <a>каталог</a> RU passage</p>` under `<html lang="uk">`,
`core/lang-part-unmarked` emits two findings equal in *every* field — rule,
verdict, grounding, scope, subject, evidence, summary, denominator — so the
report reads as one finding printed twice and a reader cannot tell that two
passages are meant at all. One level in, `core/content-language-mixed` puts the
same `node` ref in its evidence list twice: a finding that claims to cite two
passages while quoting one location N times.

**Stale citation:** the issue points at `src/collect/digest.ts:399`. That line
now lives in `src/collect/digest-dom.ts:485` (the DOM tier was split out so the
host app can share it); `digest.ts` is 89 lines of jsdom parse plus the globals
shim.

## Before / after

Same page, `core/lang-part-unmarked` and `core/content-language-mixed`:

| | before | after |
| --- | --- | --- |
| finding 1 `subject.node` | `html > body > main > p` | `html > body > main > p :: text(1)` |
| finding 2 `subject.node` | `html > body > main > p` | `html > body > main > p :: text(2)` |
| `content-language-mixed` evidence | the same node ref, twice | two distinct node refs |
| a paragraph with one passage | `html > body > main > p` | `html > body > main > p` (unchanged) |

## The fix

`digest-dom.ts` appends ` :: text(n)` where a parent holds more than one
passage, exactly as `:nth-of-type` disambiguates one level up — and **only**
where it disambiguates, so a paragraph with a single passage keeps the path it
has always published and stays a valid CSS selector.

The ordinal counts the parent's non-blank child **text nodes**: a property of
the document, never of what this collector sampled. `MIN_SAMPLED_CHARS` and the
1500-sample cap therefore cannot move a citation, which is what lets a second
collector with different sampling gates produce the same string for the same
passage.

No rule changed. Every rule that cites a text sample —
`core/lang-part-unmarked`, `core/content-language-mixed`,
`core/content-contradicts-declaration`, `core/content-chrome-untranslated`, and
the `ua` pack's market-signal citations — reads `sample.nodePath` and gets the
disambiguated pointer for free.

## Schema version: 4 → 5

`EVIDENCE_SCHEMA_VERSION` goes to 5. It is the first bump that adds no field, so
it is worth stating why it is still the right call:

- Nothing recomputes a stored bundle's paths, so a **v4 bundle replays exactly
  as before** — same ambiguous parent paths it was collected with, no rule
  branches on the shape, and `isInsideCodeElement` (the only kernel code that
  parses these strings) is unaffected either way. Nothing silently mis-cites.
- What the version buys is the question a consumer can now answer: *do two
  samples with equal `nodePath` mean one passage or several?* v4 says "possibly
  several", v5 says "one". Leaving both contracts under one version number is
  the one thing a version stamp exists to prevent.

## Costs, disclosed

- A path carrying the suffix is no longer pasteable into `querySelector`. The
  suffix appears only where the passage was ambiguous, so unambiguous citations
  keep that affordance.
- Inserting an inline element into a paragraph renumbers that paragraph's
  passages — the same churn `:nth-of-type` has always had one level up.
- Any bundle stored before v5 keeps its ambiguous citations. They cannot be
  repaired without the document, and re-deriving them would be inference.

## How the regression test proves it

`src/collect/digest.test.ts` — three tests digest real HTML and assert the
**published finding** through `CORE_RULESET` / `evaluate`, not the digest's
internal output, because that is where the harm lands:

- `gives two passages under one element two distinct citations`
- `does not publish two passages as two byte-identical findings`
- `cites two passages inside one finding at two locations`

On `main` all three fail:

```
× gives two passages under one element two distinct citations
    AssertionError: expected 1 to be 2 // Object.is equality
× does not publish two passages as two byte-identical findings
    AssertionError: expected '{"rule":"core/lang-part-unmarked","ve…' not to be
                             '{"rule":"core/lang-part-unmarked","ve…'
× cites two passages inside one finding at two locations
    AssertionError: expected 1 to be 2 // Object.is equality
```

Five more tests pin the format itself, since it is the spec a second collector
has to match: the ordinal appears only where it disambiguates, it counts a
passage the sampler dropped as too short, whitespace between block elements is
not a passage, and a `<code>` block holding two passages is still excluded from
classification through the new suffix.

## Verification

- `pnpm --filter @movar/audit test` — 853 passed (25 files); 845 on `main`, 8
  added here.
- `pnpm validate` — green, including the `audit:site` dogfood gate over
  `apps/marketing/dist`: no new findings, no stale suppression, the one existing
  entry still matched.
- `pnpm test:coverage` — statements 95.62 %, branches 92.15 %, functions
  94.79 %, lines 97.24 %. `pnpm gen:readme --refresh` moved the LOC snapshot
  only.
- `pnpm metrics:audit` — clean; `sampleTextNodes` was restructured (two
  extracted helpers) rather than suppressed to stay inside the complexity
  budget.

## Out of scope

`apps/safari-host-app/src/audit/collect.ts` — the second collector across the
process boundary — builds **no** node paths of its own: it parses with
`DOMParser` and delegates to this same `digestFromDocument`, so it picks the fix
up with the package and needs no change here.

Noticed while reading the digest, **not** fixed here and **not** covered by any
open or closed issue I could find: `headTextsOf` takes the page `<title>` with
`doc.querySelector('title')`, which is document-wide. On a page whose head
declares no title, an `<svg><title>Shopping cart icon</title>` in the body is
digested as `field: 'title'`, so `core/title-contradicts-declaration` would
adjudicate an icon's accessible name as the page's title. Worth its own issue.


rejifald:
## Review: one blocker, everything else verified green

I reproduced every substantive claim in this PR independently. The fix is real, the
stronger re-characterisation is **true**, the tests are load-bearing, and the schema
bump is justified. Details below, then the one thing that needs another pass.

### Blocking: `scripts/readme-metrics.snapshot.json` does not reproduce

The committed `loc` is **3 lines short** of what the generator produces on this tree.

```
$ npx tsx scripts/gen-readme-metrics.mts --refresh
$ git diff scripts/readme-metrics.snapshot.json
-  "loc": 61128,
+  "loc": 61131,
```

The arithmetic confirms it independently rather than blaming my machine:

| | |
| --- | --- |
| `main`'s committed `loc` | 61002 |
| this PR's net line delta in scanned files (`digest-dom.ts` +123/−20, `evidence.ts` +32/−6; `digest.test.ts` excluded by `isTest`) | **+129** |
| therefore the true value | **61131** |
| committed here | 61128 (= 61002 + 126) |

So `main`'s snapshot was exact and this one is 3 low — the `--refresh` was evidently
run three lines before the final state of the source. Reproduced twice: once with the
real script above, once with an independent re-implementation of `scanSourceStats`,
both 61131. `eslint: 46` / `fallow: 23` are correct and unchanged.

**Impact is cosmetic but real.** `(61131/1000).toFixed(1)` and `(61128/1000).toFixed(1)`
are both `61.1k`, so the rendered README is byte-identical and `pnpm check:readme`
passes (green locally, and `verify` is green in CI) — `loc` is a DYNAMIC metric the
guard reads from the snapshot rather than recomputing, so nothing will ever catch this.
It just sits in a committed source-of-truth file being wrong until someone next runs
`pnpm metrics`.

Fix: set `"loc": 61131`, or re-run `pnpm gen:readme --refresh` on the final tree.
That is the whole blocker.

### Verified: the stronger re-characterisation is real

Reproduced on `main` (reverting only `digest-dom.ts` via `git show origin/main:F > F`,
so nothing was staged), adjudicating through `CORE_RULESET`:

- **`core/lang-part-unmarked` publishes two byte-identical findings.** Not merely equal
  `subject` and `evidence` as #447 says — `JSON.stringify(findings[0]) === JSON.stringify(findings[1])`
  is literally `true` on `main` and `false` here. Every field: rule, verdict, grounding,
  scope, subject, evidence, summary, via, denominator. The issue understated it and this
  PR is right to pin the stronger form.
- **`core/content-language-mixed` cites one location twice** inside a single finding:
  `evidence node refs: ["html > body > main > p","html > body > main > p"] distinct=1`
  on `main`, two distinct refs here.
- **Ordinary prose triggers it.** `<p>RU… <a>каталог</a> RU…</p>` reproduces both shapes,
  not just the issue's `<br>` fixture. I checked both; both reproduce.

Six of the eight new tests fail on `main` with exactly the quoted messages
(`expected 1 to be 2`, and the `not.toBe` on two identical serialised findings). The two
that pass on `main` are the no-churn guards, which is correct — they pin that the fix
does *not* move an unambiguous path.

### Verified: the five format-spec tests are load-bearing

Mutation-tested, each mutation applied to `digest-dom.ts` alone:

| mutation | tests killed |
| --- | --- |
| suffix removed (= `main`) | `numbers the passages…`, `counts a passage the sampler dropped as short`, `keeps a code element visible…` |
| `position.total < 2` → `< 1` (always suffix) | `leaves a single passage the path it has always published`, `does not count the whitespace between block elements` — **and only those two** |
| `advance()` moved below the `MIN_SAMPLED_CHARS` / `isSkipped` gates | `counts a passage the sampler dropped as short` — **and only that one** |
| `isPassage` admits whitespace | `does not count the whitespace between block elements` — **and only that one** |
| suffix applied to the grandparent's path | all five |

Three of the five are uniquely pinned by a single mutation. None merely restates the
implementation. One honest note: the `isInsideCodeElement(path)` half of
`keeps a code element visible through the suffix` cannot be killed independently — since
`PATH_STEP_SEPARATOR` splits on `[\s>+~]+`, *any* appended suffix leaves the `code` step
intact. It is a guard against a future format choice rather than a mutant-killing
assertion. That is fine and worth keeping; noting it so nobody over-reads it.

### Verified: the schema bump

**Ruling: a field-less bump is justified here — inert to code, load-bearing as a wire
contract. The PR's framing is accurate and does not overclaim.**

- **Nothing reads it.** Zero matches repo-wide for any comparison or branch on
  `schemaVersion` / `EVIDENCE_SCHEMA_VERSION`. It is stamped (`collect/node.ts`,
  `safari-host-app/src/audit/collect.ts`), copied onto the report (`evaluate.ts:144`)
  and displayed (`artifact.ts:327`). Rules degrade by testing **field presence**
  (`head === undefined`, `textSampling === undefined`); the `schemaVersion 1` strings in
  `notApplicableReason` are hardcoded literals, not read from the value. The earlier
  review in this series still holds.
- **Empirically inert.** Same bundle stamped v4 and v5, `evaluate()` over both:
  40 rule results, **0 differing** — verdict, grounding, finding count, every finding's
  verdict / subject.node / summary.
- **No silent mis-citation.** A genuine v4-shaped bundle (two samples both carrying the
  parent path) replays with exactly those ambiguous citations. Nothing recomputes stored
  paths.

So the bump communicates nothing to code, and a stored v4 bundle is indistinguishable
from a v5 one at runtime. That makes it **correct-but-inert at runtime** — but *not*
useless, and *not* misleading:

- It cannot mislead, because misleading requires a consumer that branches on it and
  thereby does the wrong thing. Nothing branches, and nothing recomputes.
- It is load-bearing for the format, which is the thing `Evidence` actually is. The
  change alters the *meaning* of an existing field's values without changing its type,
  and a v5 bundle whose pages happen to have no ambiguous parents is byte-identical in
  shape to a v4 one — so a consumer **cannot** sniff which contract it is holding. The
  version is the only discriminator there will ever be. Leaving both semantics under
  version 4 would be unrecoverable.

One standing observation, pre-existing and not introduced here: because nothing
validates or compares the constant, this discipline is entirely honour-system — as true
at v2/v3/v4 as at v5.

### Verified: the ordinal is sampling-independent

This is the #438/#439 property, so I tested it against a genuinely capped page rather
than trusting the argument:

| page | `cappedAt` | `examined` / `sampled` | the target paragraph's citations |
| --- | --- | --- | --- |
| small | — | 2 / 2 | `… > p :: text(1)`, `… > p :: text(2)` |
| + 2000 filler passages after | 1500 | 2002 / 1500 | **identical** |

Truncation does not move what the report quotes. A passage dropped by
`MIN_SAMPLED_CHARS` still occupies its ordinal (`<p>x<br>long…</p>` → `text(2)`), so the
kept citation is document-truthful. Samples past the cap are dropped entirely rather
than mis-cited, which is the safe direction.

### Verified: no invariant violation, no scope creep

- `pnpm metrics:audit` → `✓ No issues in 8 changed files`. The `sampleTextNodes`
  restructure genuinely stays inside the complexity budget; **no** suppression of any
  kind was added (`eslint-disable`, `fallow-ignore`, `@ts-*`, coverage ignores — the diff
  adds none, and the snapshot's `46 / 23` counts are unchanged and correct).
- **Restructure is behaviour-preserving.** I digested a fixture exercising every gate
  (whitespace-only nodes, 1-char nodes, `script`/`style`/`noscript`/`template`, collapsible
  whitespace, `>MAX_TEXT_NODE_CHARS`, nbsp, regions, `lang`) under `main` and under this
  branch and diffed the full `DocumentEvidence` with the ` :: text(n)` suffix stripped:
  **byte-identical** — sampling report, sample count, texts, `inheritedLang`, `region`,
  `langAttributes`, `links`, `head`. The only raw-path differences are the five passages
  whose parent genuinely holds more than one.
- Kernel purity and the grading law hold (87 tests): `digest-dom.ts` stays browser-safe
  under `collect/`, `evidence.ts` stays pure, `classified` cannot be written as failing.
- **Nothing keys on `nodePath` string equality.** `suppress.ts` `matches()` compares
  `finding.rule` and `finding.subject.path` / `.url` only — never `subject.node`.
  `apps/marketing/audit-suppressions.json`'s single entry is
  `core/inventory-varies-across-pages`, which cites no node. Across the tree `nodePath`
  is only ever passed to `nodeRef()` or `isInsideCodeElement()`; no `Map`/`Set` key, no
  equality test, no grouping. No suppression can go stale from this change.
- Safari claim holds: `apps/safari-host-app/src/audit/collect.ts` builds no paths — it
  is `digestFromDocument(new DOMParser().parseFromString(html, 'text/html'), options)`
  and inherits the fix.
- Scope is exactly the seven expected files; `git diff origin/main --summary` is empty
  once the local `pnpm install` chmod on `collect/cli.ts` is stripped. Test count 853
  (25 files) reproduces exactly.

### On the two disclosed costs

Both are honestly disclosed and acceptable. The `querySelector` cost is bounded by
construction — the suffix appears only where the passage was ambiguous, and an ambiguous
passage never had a pasteable selector *that identified it* in the first place, so
nothing that worked before stops working. The renumbering cost is the same churn
`:nth-of-type` already has one level up, and the report is a point-in-time artifact
about a point-in-time page.

### Out of scope, agreed

The `headTextsOf` / `doc.querySelector('title')` observation is a real finding and
correctly left out of this PR. Worth its own issue.

---

Everything above passes. Only the 3-line `loc` snapshot needs another pass; push that
and I will re-run the gate.

PR #488, opened 2026-08-15
fix(metrics): exclude generated dirs from the source-line count

rejifald:
## What

`scanSourceStats()` in `scripts/gen-readme-metrics.mts` skipped build directories
by name, but not `apps/marketing/.astro`. `astro sync` — which the marketing
package's own `lint` script runs, as do `astro build` and `astro check` — writes
a ~208-line `content.d.ts` and a `types.d.ts` there. Both end in `.ts` and
neither looks like a test, so both were counted as source.

So `loc` depended on whether a build or a lint had run before `pnpm metrics`,
not on the source tree:

| | `loc` | README badge |
| --- | --- | --- |
| before, `.astro` present | 61127 | `61.1k` |
| after, `.astro` present | 60917 | `60.9k` |
| after, clean checkout | 60917 | `60.9k` |

That is why historical snapshots drifted unpredictably (+206 at #406, +469 at
#424, +176 at #427) and why the badge recently rendered `56.9k` where the true
source count was `56.7k`.

## Changes

**`BUILD_DIRS` gains the generated dirs that can hold `.ts`** — `.astro`,
`.wrangler`, `storybook-static`, `test-results`, `demo-results`. The
`playwright-report*` family (`playwright-report`, `-live`, `-compare`, all
written by the e2e configs) is matched by prefix via a new `isBuildDir()` helper,
since `.gitignore` covers them with a glob and an exact-name `Set` can't. Checked
that no tracked file lives under any of those names, so nothing real is skipped.

**The metrics gate now pins `loc` alongside coverage.** It is a pure function of
the source tree once the generated dirs are excluded, and the gate's `--refresh`
step already recomputes it, so the check costs nothing — leaving it unpinned is
why the drift went unnoticed for so long. It is folded into the existing
freshness check: exit 2, non-waivable, with coverage and LOC reported together so
one `pnpm metrics` run fixes both. The failure message names `BUILD_DIRS` as the
likely culprit when the number is reproducible locally but not in CI.

The two halves are coupled: without the `.astro` exclusion, pinning `loc` would
have failed the gate for anyone who ran `pnpm lint` before `pnpm metrics`.

Docs (`docs/metrics-gate.md`) and the workflow header updated to match. The gate
test gains two cases — stale LOC fails with exit 2, and still fails with the
`accept-metrics-regression` label — alongside the three existing floor cases.

## For the reviewer

- **No snapshot or README change.** This branch's committed `loc` of 60917 was
  already the clean value, so the fix locks it in rather than correcting it.
- **I ran `gen:readme --refresh`, not the full `pnpm metrics`.** The heavy half
  (fallow, coverage, the content-bundle guard) cannot move any number here — the
  diff touches only `scripts/`, `docs/`, and `.github/` — and refreshing `health`
  would have added churn the docs themselves describe as drifting commit-to-commit
  with git history. Happy to run the full refresh if you'd rather.
- **Verified empirically**, not just by reading: ran `astro sync` to generate the
  real 208 lines, re-ran the refresh, and got a byte-identical snapshot and
  README. Also swept a checkout carrying `.astro`, `playwright-report`, and
  `test-results` — zero gitignored directories contribute source lines, and it
  agrees with a clean worktree exactly.
- **No changeset**, since this ships no package code — it is build tooling, CI
  config, and docs. Say the word if you want one anyway.
- `packages/audit/**` is untouched, as requested. Worth knowing separately:
  `pnpm install` in a fresh worktree flips `packages/audit/src/collect/cli.ts`
  from mode 644 to 755 (pnpm chmod'ing a `bin` target). I reverted it here, but it
  will keep reappearing until the file is committed as 755.
- **Unrelated snag hit while pushing**, noted in case it bites you too: the
  `pre-push` e2e step failed every popup/options spec in this fresh worktree
  because `extension:build:e2e` is `"cache": true` with no `outputs` declared, so
  nx replayed a cache hit without materialising `apps/extension/.output/`.
  `nx run extension:build:e2e --skip-nx-cache` fixed it and the suite went 21/21.
  Same gap on `extension:build`, `diagnostics:build`, and `marketing:build`.
  Tracked separately — not touched here.
- **These new test cases do not run in CI.** `pnpm test:metrics-gate` — like
  `test:promises`, `test:release-notes`, `test:action-pins`, and `test:edge-poll`
  — is not referenced by any workflow or by `lefthook.yml`. Pre-existing gap,
  tracked separately.

### DIFF
--- a/.github/workflows/metrics-gate.yml
+++ b/.github/workflows/metrics-gate.yml
@@ -4,10 +4,12 @@ name: Metrics gate
 # maintainer explicitly accepts the regression. Three things are checked
 # (see scripts/metrics-gate.mts for the full rationale):
 #
-#   1. Coverage freshness — the committed `readme-metrics.snapshot.json` must
-#      match freshly-recomputed coverage. Keeps the README badge honest AND
-#      keeps `main`'s snapshot trustworthy as the regression baseline below, so
-#      we never have to run coverage on the base commit.
+#   1. Snapshot freshness — the committed `readme-metrics.snapshot.json` must
+#      match the freshly-recomputed coverage AND source-line count. Keeps the
+#      README badges honest AND keeps `main`'s snapshot trustworthy as the
+#      regression baseline below, so we never have to run coverage on the base
+#      commit. (LOC needs no extra step here: the `--refresh` below recomputes
+#      it unconditionally.)
 #   2. Coverage regression — recomputed coverage must not drop below the base
 #      commit's snapshot.
 #   3. Code-quality regression — `fallow audit --base <base>` must not introduce
--- a/docs/metrics-gate.md
+++ b/docs/metrics-gate.md
@@ -11,11 +11,11 @@ but can't notice when the snapshot itself drifts from reality. This gate does.
 Implemented in [`scripts/metrics-gate.mts`](../scripts/metrics-gate.mts), run by
 [`.github/workflows/metrics-gate.yml`](../.github/workflows/metrics-gate.yml) on
 every PR. It recomputes the dynamic metrics for the PR head and compares them
-three ways:
+four ways:
 
 | #   | Check                       | Fails when                                                                   | Overridable?                                  |
 | --- | --------------------------- | ---------------------------------------------------------------------------- | --------------------------------------------- |
-| 1   | **Coverage freshness**      | the committed `readme-metrics.snapshot.json` coverage ≠ recomputed coverage  | No — a wrong number is wrong, not a trade-off |
+| 1   | **Snapshot freshness**      | the committed `readme-metrics.snapshot.json` coverage or LOC ≠ recomputed    | No — a wrong number is wrong, not a trade-off |
 | 2   | **Coverage regression**     | recomputed line/branch coverage drops below the base commit's snapshot       | Yes (label)                                   |
 | 3   | **Code-quality regression** | `fallow audit --base <base>` finds new dead code, complexity, or duplication | Yes (label)                                   |
 | 4   | **Coverage floor**          | recomputed line/branch coverage drops below the absolute `COVERAGE_FLOOR`    | No — a waivable floor is the ratchet it stops |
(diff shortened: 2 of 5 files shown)


rejifald:
## Review — REQUEST_CHANGES (`.astro` half now superseded by #506)

The bug was real and was reproduced on a live dev checkout carrying `apps/marketing/.astro`, `apps/e2e/playwright-report` and `apps/e2e/test-results`: old `BUILD_DIRS` → **61333**, new → **61131**; on a clean worktree at the same commit both → 61131. Snapshot needed no update — main's committed `loc` already equalled the clean scan, so the gate did not go slack. Gate tests 5/5 locally, and `metrics-gate` passed in CI on head `c97124e`.

**#506 (`4c837e2`) has since landed the `.astro` exclusion**, extracted the walk into `scripts/lib/source-stats.mts`, and added a live tripwire asserting the scan visits no git-ignored path and drops no git-known file. `git merge-tree` reports a conflict in `scripts/gen-readme-metrics.mts` because #506 deletes the block this PR expands.

### What remains valuable here
Verified against main after #506: **five of this PR's skip-list entries are still absent** — `test-results`, `demo-results`, `storybook-static`, `.wrangler`, `playwright-report`. Those are worth re-landing.

### What should not be carried across
**`scripts/gen-readme-metrics.mts:230` — `name.startsWith('playwright-report')` over-captures.** Verified on a synthetic tree that it silently skips `packages/playwright-reporter/src/real-source.ts` — the exact "real source hidden from the gate forever" failure mode. No such directory exists today, which is also why #506's tripwire would *not* catch it. Use `name === 'playwright-report' || name.startsWith('playwright-report-')`.

**`scripts/metrics-gate.mts:180` — the exact `loc` pin.** `loc` moved on all 8 recent snapshot commits (60608→…→61131). The documented remedy is `pnpm metrics`, which re-runs the full coverage suite — minutes — after any fixup touching a single TS line. #489's tolerance band is the better policy.

Also noted: `.gitignore` names seven further directories absent from `BUILD_DIRS`; none holds a `.ts` today, but pinning `loc` exactly turns that soft invariant into a hard CI failure the moment it slips. Driving the scan off `git ls-files` would retire the class.

### On the red CI
`ci-gate` is a pure aggregator that failed **only** because `e2e-offline` did (130 passed / 1 failed — `apps/e2e/src/offline/content-script.spec.ts:362` timing out at 5s under `retries: 0`). This PR touches only `.github/`, `docs/` and `scripts/`, and the last 8 `main` runs were green — so that failure was flake needing a re-run, not a code fix.

Planned: a single combined PR taking #489's gating policy plus the five entries above (without the prefix matcher), after which this and #489 can close.

*Automated review.*

rejifald:
Closed automatically by #511 (`07cbc7a`) — recording what carried across and what did not.

**Carried across:** the five skip-list entries main was still missing — `test-results`, `playwright-report`, `storybook-static`, `demo-results`, `.wrangler`. (The `.astro` entry, the one your measurement was actually about, had already landed via #506.)

**Deliberately not carried across:**

- `name.startsWith('playwright-report')` — verified on a synthetic tree that this silently skips `packages/playwright-reporter/src/real-source.ts`, the "real source hidden from the gate forever" failure mode. #506's live tripwire would not catch it either, because no such package exists to be shadowed. #511 matches the exact name plus a separate `playwright-report-` prefix list, covering `-live` and `-compare` and nothing a package would plausibly be called.
- The exact `loc` pin. `loc` moved on all 8 recent snapshot commits, and the documented remedy was `pnpm metrics` — a full coverage run — after any fixup touching one TypeScript line. #511 uses a ±500 band plus `gen:readme --refresh-source`, which re-pins from a rescan in ~0.5s. The band is what makes the check affordable enough to be non-waivable.

Your measurement was the load-bearing part: old `BUILD_DIRS` → 61333, new → 61131 on a dirty tree, both → 61131 clean. That established the bug was real and bounded.

For the record, the red CI here was never this PR's fault — `ci-gate` is a pure aggregator that failed only because `e2e-offline` did (130 passed / 1 failed, a 5s timeout under `retries: 0`).

PR #489, opened 2026-08-15
fix(metrics): gate loc and suppressions against a live scan, not the snapshot

rejifald:
## The hole

`docs/metrics-gate.md` check 1 recomputes **coverage only**. `check:readme` +
`gen-readme-metrics.mts --check` verify that the README block renders consistently
with the *committed snapshot* — by design they "recompute the badge block without
running any tool". So nothing anywhere compared `loc` against reality, and any PR
that changed TypeScript under `apps/`/`packages/` without running `pnpm metrics`
left the number stale indefinitely.

`main` once understated its own line count by 184 lines (`56.7k` where the truth
was `56.9k`). The only thing that corrected it was #463 refreshing the snapshot
for an unrelated coverage reason. That's luck, not a process. The same gap applied
to the `eslint`/`fallow` suppression counts from the same `scanSourceStats()` pass.

## Measured before deciding

The last 39 consecutive snapshot commits on `main`:

| metric | changed per PR |
| --- | --- |
| `loc` | **39/39** |
| coverage | 19/39 |
| `suppressions` | 5/39 |

Also: **30 of the last 30** commits touched the snapshot *and*
`docs/REFACTORING-QUEUE.md`. So the snapshot was already refreshed on nearly every
PR — the drift wasn't "nobody refreshes", it was "the refresh is a point-in-time
local run that nothing verifies", so any post-refresh edit silently invalidated it.

That asymmetry is what drives the design below.

## What changed

**Check 5 in `metrics-gate.mts`, non-waivable** — `loc` within a **500-line band**,
`suppressions` **exact**.

- `loc` responds to every added or deleted line, so an exact gate leaves zero
  slack: refresh, push, take one review comment, red again. 500 lines covers the
  median PR (67) many times over.
- The error is now **bounded and non-accumulating** — every PR is compared against
  a live scan of *its own head*, so a stale figure inherited from `main` buys the
  next PR nothing. That's the whole difference from unbounded silent drift.
- Suppressions are small integers printed verbatim, moved 5 times in 39 commits,
  and adding an `eslint-disable` is exactly the change that should cost something.

**It costs no CI time.** The workflow already runs `gen:readme --refresh`, which
rescans the tree unconditionally — the gate was computing the truth and discarding it.

**The badge and the band are a matched pair.** `loc` now renders `~61k` instead of
`60.9k`, because ±500 lines is all the gate guarantees; a tenths digit claimed ten
times that precision. Documented in three places that widening `LOC_TOLERANCE`
without coarsening the render re-opens the hole. This is why the original 184-line
case is now doubly handled: bounded by the gate, and invisible at the rendered
resolution.

**New `--refresh-source` mode** rescans only LOC + suppressions — measured at
**0.4s** vs minutes for `pnpm metrics` — and, unlike a bare `--refresh`, cannot
overwrite good coverage with a stale local `coverage/`. Exit `4` points at it;
exit `2` still means "run `pnpm metrics`".

**Queue churn**, addressed at the same time: `docs/REFACTORING-QUEUE.md` was touched
by all 30 of the last 30 commits, usually only by two advancing timestamps that
contributors reverted by hand. The generator now skips the write when nothing
substantive changed, and the file joins `.prettierignore` — without that, prettier's
table realignment restarts the edit war on the next run (same precedent as
`packages/theme/styles`).

## Verification

- 11/11 gate tests pass (8 new): band boundary at exactly ±500, past-band → exit 4,
  accept label does **not** rescue it, suppression off-by-one → exit 4, stale
  coverage + stale source → exit 2 (not 4).
- End-to-end against the real snapshot: 184-line drift → exit 0; 4200-line drift →
  exit 4 with the cheap-refresh message.
- `check:readme`, `check:suppressions`, `check:action-pins`, `lint:root`,
  `format:check` green; full pre-push (build + e2e + `metrics:audit`) green.

## Reviewer notes

- **Snapshot values are untouched.** I verified the committed `loc` (60917) and
  suppressions (46/23) against a live scan — both accurate — so only `_comment`
  changes. This deliberately does not "re-run `pnpm metrics` and commit".
- **Branch is 3 commits behind `main`** (`185a896`, `cc04878`, `0704e8e`), which
  landed during this work. `0704e8e` touched the snapshot (`loc` is now 61002), so
  expect a trivial `_comment` conflict on update. `61002` still renders `~61k`, so
  no README change follows.
- The first `pnpm metrics` run after this merges will rewrite `REFACTORING-QUEUE.md`
  once to un-pad the tables prettier previously aligned. Expected, one-time, then stable.
- Unrelated, not addressed: `check:suppressions` reports 30 while the badge says 23 —
  different scopes (repo-wide genuine directives vs. `fallow-ignore` substrings in
  `apps`/`packages` `.ts`/`.tsx`). Both correct for what they measure.


rejifald:
## Review — REQUEST_CHANGES

The core design is sound and was checked rather than assumed: the workflow's pre-existing `cp` step stashes the committed snapshot **before** `--refresh` overwrites it, and `LOC_TOLERANCE` is a hardcoded 500 — so the gate compares a value frozen at commit time against a fresh scan. **Not a tautology.** Performance is a non-issue: `scanSourceStats()` runs in 42–150 ms over 405 files, one gate invocation is 0.57 s, `--refresh-source` 0.40 s, and `refreshSnapshot()` already called that function before this PR. Test suite 11/11.

### Blocker

**No CI has ever run on this PR.** `refs/pull/489/merge` does not exist because the branch conflicts with main, and both workflows are `on: pull_request` only, so GitHub never created a run — `gh run list --branch` is empty and `check-runs` totals 0. Sibling #488 at the same merge-base has a full check set, which isolates the conflict as the cause. Since `metrics-gate` is a required check and **this PR is that gate**, it must be rebased and observed green before merge.

### Should-fix

**`scripts/gen-readme-metrics.mts:199` — the scan this PR elevates to gate authority was not deterministic.** Two clean checkouts of the identical commit `326133f` scanned **61131** vs **61333** — a 202-line gap from `apps/marketing/.astro/content.d.ts` + `types.d.ts`, written by `astro sync` via marketing's `lint`/`typecheck`/`build`. CI never runs astro sync in this job, so CI always saw the clean number while any developer with a built worktree committed clean+202 — silently consuming ~40% of the ±500 band this PR's "bounded and non-accumulating" argument rests on.

**This is now fixed on main by #506 (`4c837e2`)**, which also extracted the walk into `scripts/lib/source-stats.mts` — so `--refresh-source` will need to call `scanSourceStats(repoRoot)` from the new module after rebasing.

**`scripts/metrics-gate.mts:225`** — the same sensitivity applies to `suppressions`, gated *exactly* with no band and no waiver. Harmless today (verified 46/23 in both a clean and a built tree), but a future generated `.ts` carrying `eslint-disable` would produce a non-waivable CI failure unreproducible locally.

### Nits
- `gen-readme-metrics.mts:452` — `~${Math.round(loc/1000)}k` under a ±500 band does not make drift invisible: truth 61,400 renders `~61k` while a gate-legal 61,900 renders `~62k`.
- `metrics-gate.mts:212`/`:225` — check 5 silently skips when recomputed values are absent, while check 1 hard-throws at `:179`.
- `scripts/fallow-targets.mts:29` — the REFACTORING-QUEUE churn suppression is an independent change bundled in. (It is a genuine annoyance: running `pnpm metrics` for any reason dirties that file.)
- PR body's reviewer notes are stale — the real merge-base is `f8a7609` (5 commits behind, not 3) and main's committed `loc` was 61131, not the predicted 61002.

Planned: a single combined PR taking this PR's gating policy plus #488's five remaining `BUILD_DIRS` entries (without its over-capturing `playwright-report` prefix matcher), after which this and #488 can close.

*Automated review.*

rejifald:
Closed automatically by #511 (`07cbc7a`), which is essentially this PR's design rebuilt on `scripts/lib/source-stats.mts` after #506 moved the scan.

**Carried across intact:** the ±500 tolerance band for `loc`, exact gating of the suppression counts, a distinct non-waivable exit code (4), and the fast `gen:readme --refresh-source` remedy. Your central argument — that the workflow stashes the committed snapshot before `--refresh` overwrites it, so the gate compares a value frozen at commit time against a fresh scan and is therefore not a tautology — is why the check has that shape.

**Changed:** `--refresh-source` calls `scanSourceStats(repoRoot)` from #506's module rather than the old in-script function, and the README badge keeps its existing precision rather than being coarsened to `~61k`. Under a ±500 band that rounding can still be one unit off in its only significant digit (truth 61,400 → `~61k`, gate-legal 61,900 → `~62k`), so it did not buy what it claimed.

**Not carried across:** the `fallow-targets` / `.prettierignore` change suppressing `docs/REFACTORING-QUEUE.md` timestamp churn. It is a real papercut — I hit it three separate times running `pnpm metrics` for other PRs — but bundling it into a gating PR was the scope objection raised in review, and repeating that would have been inconsistent. It deserves its own PR and I would support one.

Your `.astro` measurement (two clean checkouts of `326133f` scanning 61131 vs 61333) was independently reproduced and drove #506, which had to land first: a live-scan gate over an unfixed walk would fail on any developer tree that ran `astro check`.

The blocker was never the code — this branch conflicted with main, so GitHub never created `refs/pull/489/merge` and zero CI ever ran on it.

PR #490, opened 2026-08-15
fix(audit): take the served-language denominator from the population, not the sample

rejifald:
Half of #483 — the `served-language.ts` call site. The `ua.ts` half is deliberately
left out and is described under [What this does **not** fix](#what-this-does-not-fix).

## The defect

`classifiedPageLanguage` in `packages/audit/src/served-language.ts` built its own
`Denominator` instead of calling the shared helper:

```ts
const samples = page.document.textNodes;
…
denominator: { examined: samples.length, matched }
```

`textNodes` is a **sample**. The collector caps body sampling at
`MAX_TEXT_NODE_SAMPLES` (1500) and records what the walker actually saw in
`textSampling.examined` (`schemaVersion` 3, #475). So a truncated page quoted the
cap as if it were the page — understating `M` by the whole truncation and
inflating the share the finding publishes, in the direction of the accusation.

This is the third instance of the class `packages/audit/AGENTS.md` already records
(_"`textNodes` is a sample; `textSampling.examined` is the population"_), after
#478's volume comparison and the five call sites #438 migrated.

## Blast radius

Wider than the other two, because this one is **kernel** rather than a single
rule. `classifiedPageLanguage` is the classifier half of the hybrid seam, so it
feeds both families that read it:

- `core/switch-no-effect` (`rules/switch.ts`), which quotes the numbers in its summary
- family C's hybrid `core/serving-*` findings (`rules/serving.ts`), via `mergedDenominator`

And `mergedDenominator` **sums** these across determinations, so the understatement
compounds instead of staying per-page:

| | before | after |
| --- | --- | --- |
| one page, 4000 examined / 2 sampled | `2 of 2` | `2 of 4000` |
| two such pages, merged | `4 of 4` | `4 of 8000` |
| pre-v3 bundle, 2 sampled | `2 of 2` | `2 of 2` (unchanged) |

`2 of 2` is not a smear in the sense `finding.ts` warns about — it is worse. It
reads as a total, unanimous verdict on a page the audit looked at 0.05 % of.

## The fix

One line: `textNodeDenominator(page, matched)`, the helper the other callers
already use. Two properties worth checking in review:

- **Exclusions never narrow `examined`.** Nothing filters on the way to the
  denominator here — a node the classifier abstained on stays in `M` and only
  fails to raise `N`. That is what keeps the share honest rather than flattering
  to whichever verdict won.
- **Old bundles still replay.** A bundle written before `schemaVersion` 3 carries
  no counts, so it falls back to the sample length — exactly the denominator it
  was originally published with.

## Wording

`core/switch-no-effect` said _"N of M sampled text nodes"_. That was an accurate
label for the wrong denominator, and would be the wrong label for the right one:
in this repo "sampled text nodes" means the denominator **is** the sample, which
is why #438 dropped it from the five it migrated. All three existing
`textNodeDenominator` callers (`content-language.ts`, `page-declaration.ts`,
`inventory-hreflang.ts`) say plain "text nodes"; this one now does too.

`serving.ts` needed no change — it never quotes the counts inline. Its only
hybrid phrasing is `viaNote`'s _"served language read from sampled text"_, which
describes what the classifier read rather than what the denominator counts, and
stays true. `artifact.ts` renders the structured field as _"N of M examined"_,
also neutral.

## Tests

New `src/served-language.test.ts` (12 tests) pins the kernel contract directly,
since the module had no test file of its own: truncated page quotes the
population, abstention does not narrow it, a pre-v3 bundle falls back, and
`mergedDenominator` sums populations (8000, not 4).

Two rule-level regressions in `rules/switch.test.ts` on top, because the
**published summary** is what a site is confronted with, not the structured
field — a kernel-only test would pass while the sentence still said `2 of 2`.

Both use the `capped(examined, sampled)` helper pattern from `ua.test.ts`.

Reverting the one-line fix fails five of the new assertions; the pre-v3 and
abstention cases pass either way, which is the point of them.

## What this does not fix

- **The `ua.ts` half of #483.** `classifyDefaultLanguage` (`rules/ua.ts:511`)
  still returns `examined: nodes.length`, and narrows twice — the sample, then a
  blank-node `.filter()`. It lands harder than this one (`ua/state-language-not-default`
  is statute-cited and names a company), so it is getting its own change rather
  than riding along here. #483 stays open until it lands.
- **#435** — that `classifiedPageLanguage` classifies raw node text rather than
  routing through the shared `classifySamples` gates. Untouched here on purpose:
  it changes which nodes are classified, this changes only what they are counted
  against.

## Verification

- `pnpm --filter @movar/audit exec vitest run` — 26 files, 860 passed
- `pnpm --filter @movar/audit exec tsc --noEmit` — clean
- pre-commit (eslint, prettier, suppressions, typecheck, test) and pre-push
  (build, `metrics:audit`) both green

### DIFF
--- a/packages/audit/src/rules/switch.test.ts
+++ b/packages/audit/src/rules/switch.test.ts
@@ -8,6 +8,7 @@ import type {
   PickerOption,
   RedirectHop,
   TextNodeSample,
+  TextSampling,
 } from '../evidence';
 import type { RuleResult } from '../report';
 import type { Ruleset } from '../ruleset';
@@ -78,13 +79,23 @@ function targetPage(
   htmlLang: string | null,
   url = UK_PRODUCT,
   textNodes: readonly TextNodeSample[] = [],
+  textSampling?: TextSampling,
 ): PageEvidence {
   return makePage({
     id: 'page-uk',
     url,
     reach: 'declared-target',
-    document: makeDocument({ htmlLang, textNodes }),
+    document: makeDocument({
+      htmlLang,
+      textNodes,
+      ...(textSampling === undefined ? {} : { textSampling }),
+    }),
   });
+}
+
+/** What a collector writes when its sampling ceiling, not the page, ended the walk. */
+function capped(examined: number, sampled: number): TextSampling {
+  return { examined, sampled, cappedAt: sampled };
 }
 
 /** A 301 to the Ukrainian homepage, then a 302 straight back to the Russian page. */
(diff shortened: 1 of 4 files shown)


rejifald:
Superseded by #503 (`12f2a83`).

This PR fixed the `served-language.ts` half and left the `ua.ts` half to #483; #503 did both, so main now covers this strictly. Verified before closing: the old `{ examined: samples.length, matched }` pattern is gone from both files in main.

Your `mergedDenominator` coverage was not lost. The block *"sums the populations, so a truncation cannot compound across pages"* was ported onto #503 before it merged, rewritten in that file's own idioms (`pageWith`, an inline sampling literal) rather than importing a second set of helpers. It is load-bearing: reverting `served-language.ts` to main makes it fail with `expected { examined: 2, matched: 2 } to deeply equal { examined: 4000, matched: 2 }`.

Still open from this PR's review, both pre-existing and unrelated to the fix:
- `rules/switch.ts:237` publishes *"2 of 4000 text nodes classify as ru — following the switch does not change the language"* with no hedge, where the three sibling callers append `MEASUREMENT_CAVEAT`.
- `rules/inventory-hreflang.ts:391` sets `matched` to every node that classified *at all*, then says those nodes "classified ${dominant.language}" — an overstatement in the direction of the accusation.

Closed as part of an automated review sweep; reopen if you disagree.

PR #496, opened 2026-08-15
fix(audit): take the served-language denominator from the population, not the sample

rejifald:
#475 put the sampling report on the wire (issue #438) and migrated the five `textNodeDenominator` call sites. `served-language.ts`'s `classifiedPageLanguage` was missed, because it assembles a `Denominator` literal instead of calling the helper — `{ examined: samples.length, matched }`, where `samples` is `page.document.textNodes`, the 1500-node sampling cap quoted as if it were the page. That understates `M` and thereby **inflates** the share the finding publishes, in the direction of the accusation.

Because this is the hybrid seam rather than a rule, the floor did not stay one page's:

- `core/switch-no-effect` published a share of the cap as a share of the page, in both the denominator and a summary that said _"sampled text nodes"_.
- Family C's hybrid `core/serving-*` findings inherited it through `mergedDenominator`, which **sums** one denominator per response — so the understatement compounded with every page a finding cited. Two truncated responses read as _"4 of 4"_ — a unanimous verdict on both — where the walker had examined 1400 nodes between them.

## What changed

`classifiedPageLanguage` routes through `textNodeDenominator(page, matched)`. The numerator stays the count over the sample the classifier actually saw: that widest-honest pairing is the one every other classified denominator in the kernel uses, and it errs low — the safe direction for an accusation.

`core/switch-no-effect`'s summary reads its numbers off the denominator the finding carries and drops _"sampled"_, which stops being true of the wider count (`finding.ts`' canonical phrasing is _"3 of 340 text nodes"_, and the existing `textNodeDenominator` callers all say plain "text nodes"). `serving.ts` needed no wording change: its only _"sampled"_ is `viaNote`'s _"served language read from sampled text"_, which is about the reading rather than the count, and no family-C summary quotes the numbers.

A bundle written before `schemaVersion` 3 carries no counts and falls back to the sample length — exactly what this quoted before — so a stored bundle still replays to the finding it produced.

`packages/audit/AGENTS.md` § Gotchas gains the rule the defect breaks: **never assemble a `Denominator` literal**, however local it looks, plus the `mergedDenominator` compounding and the plain-"text nodes" wording convention.

## Tests

Three regressions, all driven through `evaluate()`, following `content-language.test.ts`'s `measures a truncated sample against what was examined, not what survived`:

- `core/switch-no-effect` — a truncated page (`examined 900`, `sampled 2`) must publish `{ examined: 900, matched: 2 }` and say _"2 of 900 text nodes"_.
- `core/switch-no-effect` — a bundle with no `textSampling` at all (the pre-`schemaVersion`-3 shape) falls back to the sample length, so old bundles still replay.
- `core/serving-header-partial` — two truncated responses sum to `{ examined: 1400, matched: 4 }` rather than `4 of 4`, which is the compounding a per-rule fix would have missed.

Without the source change, four assertions fail:

```
core/serving-header-partial  expected { examined: 4, matched: 4 } to deeply equal { examined: 1400, matched: 4 }
core/switch-no-effect        expected { examined: 2, matched: 2 } to deeply equal { examined:  900, matched: 2 }
```

plus the two summary assertions that no longer say "sampled".

`pnpm --filter @movar/audit exec vitest run` → 848 passed / 25 files. `tsc --noEmit`, `prettier --check .` and `pnpm check:readme` are all green, as is the repo-wide pre-commit lint/test/typecheck.

No `readme-metrics.snapshot.json` bump: `scanSourceStats` excludes tests, and this PR's +13 production lines leave the rendered `61.0k` untouched.

## Refs #483, deliberately not Closes

The issue tracks two call sites. Its other half — `ua.ts`'s `classifyDefaultLanguage`, the higher-stakes one, since `ua/state-language-not-default` is a `fail` carrying `UA_CITATION` about a named company — is **not** on `main` and is not in this branch's history. It is unmerged in #484, which carries both halves. Closing #483 from here would close it while that half still publishes _"2 of 2 sampled text nodes"_ on `main`, so this says `Refs`. Whichever of the two lands second can claim the close.

Note also that #484 fixes this half identically, and branch `claude/quizzical-nightingale-5f630f` fixes it independently — worth reconciling before merging more than one.

## On #435

Left alone deliberately. #435 (`servedLanguage` classifies text the shared text-sample gates exclude) touches the same function, but it changes *whether* a determination happens at all — routing raw `sample.text` through `classifySamples`' 40-character floor, proper-noun and `<code>` exclusions — which changes which findings fire and can unmask `core/switch-no-effect` drafts a bogus determination was suppressing. That is a different blast radius from a measurement fix. This change is forward-compatible with it and makes it smaller: once the numerator comes from `classifySamples`, it pairs with the `textNodeDenominator` population exactly as `content-language.ts` already does.


rejifald:
## Review — REQUEST_CHANGES (superseded; two fragments worth salvaging)

Superseded on every production line by #503 (`12f2a83`), which landed the same `served-language.ts` fix **and** `ua.ts:classifyDefaultLanguage` — a call site this PR never touches, making it strictly narrower.

Inventory against current main:

| Hunk | Status |
|---|---|
| `served-language.ts` `textNodeDenominator` call | already in main (`:84`) |
| `switch.ts` drop "sampled" | already in main, byte-identical (`:237`) |
| `switch.test.ts` `/2 of 2 text nodes/` | already in main (`:226`) |
| `switch.test.ts` truncated-sample test | equivalent to main `:236` |
| `switch.test.ts` pre-v3 fallback test | redundant — main `served-language.test.ts:82` |
| `serving.test.ts` `sums what each response examined…` | **genuinely absent from main** |
| `AGENTS.md` rewritten `textNodes` bullet | **genuinely absent from main** |

### Worth salvaging

**The `serving.test.ts` test is a real guard, measured not assumed.** Applied to pristine main it needs no adaptation, applies cleanly and passes 51/51. Reverting main's `served-language.ts:84` fix makes it fail with `expected { examined: 4, matched: 4 } to deeply equal { examined: 1400, matched: 4 }`, and it is the only failure in that file. It is the sole family-C fixture where `examined ≠ sampled`, pinning the composed end-to-end number at the layer that produces the user-visible finding. Tempering: main already fails 5 other tests on that same revert, so this hardens rather than rescues.

**Two of the three AGENTS.md claims are true and undocumented** — that `mergedDenominator` sums one denominator per response so a seam-level floor compounds per page cited, and that a summary quoting such a count says plain "text nodes". The second was verified against every summary in main that quotes one (`content-language.ts:158/192/253`, `page-declaration.ts:405`, `inventory-hreflang.ts:425`, `switch.ts:237`, `ua.ts:562`) — all plain; the two surviving "sampled" uses are correct and different in kind.

### Must not be lifted verbatim

**`AGENTS.md` — "never assemble a `Denominator` literal, however local it looks" is false.** `rules/content-language.ts:349` (`headDraft`) correctly assembles one whose `examined` counts *head fields*; `textNodeDenominator` would be flatly wrong there, and that function's own doc comment at `:332` says so. #484 independently wrote the same over-broad rule — two PRs converging on the same mistake suggests the correct scoping (a **body-text** denominator comes from `textNodeDenominator()`, never a literal) simply isn't written down anywhere.

Conflicts against main in `AGENTS.md`, `switch.test.ts`, `served-language.ts` — the latter two solely because #503 landed the same fix.

*Automated review.*

rejifald:
## Closing as superseded — the salvage is #539

Every production hunk here is on `main` via #503 (`12f2a83`), which landed the same `served-language.ts` fix **and** `ua.ts:classifyDefaultLanguage`, the call site this PR never touched. Re-verified against `main` at `7f6b1ee1`:

| Hunk | Disposition |
|---|---|
| `served-language.ts` `textNodeDenominator` call | on main (`:86`) |
| `switch.ts` drop "sampled" | on main, byte-identical (`:237`) |
| `switch.test.ts` `/2 of 2 text nodes/` | on main (`:226`) |
| `switch.test.ts` truncated-sample test | equivalent on main (`:236`) |
| `switch.test.ts` pre-v3 fallback test | redundant — `served-language.test.ts:82` |
| `serving.test.ts` `sums what each response examined…` | **salvaged → #539** |
| `AGENTS.md` `textNodes` bullet | **salvaged → #539, rescoped** |

Both halves of #483 are now on `main` (`ua.ts:518` takes the helper too), so nothing this branch carries is still missing from the product.

### What #539 took

The `serving.test.ts` guard, unchanged in substance: it is the only family-C fixture where `examined ≠ sampled`, and reverting `served-language.ts:86` makes it the sole failure in that file. It pins the *composed* number — `mergedDenominator` sums one denominator per determination, so the seam's floor compounds per page a finding cites, which every per-page test elsewhere misses by construction.

The `AGENTS.md` bullet, **rescoped**. This PR's *"never assemble a `Denominator` literal, however local it looks"* is false: `content-language.ts:349`'s `headDraft` correctly assembles one whose `examined` counts *head fields*, and `textNodeDenominator` would be flatly wrong there — its doc comment at `:332` says as much. #539 writes the correct rule (a **body-text** denominator comes from `textNodeDenominator()`) plus the two undocumented facts worth keeping: the `mergedDenominator` compounding, and the plain-*"text nodes"* wording convention, with `ua.ts:939`'s explicit `sampled of examined` pair exempted.

The `capped()` helper was dropped — main's sibling `switch.test.ts:236` inlines the literal, so the helper would have been a local idiom of one file.

Nothing else from this branch is worth carrying, and it conflicts against `main` in three files solely because #503 landed the same fix. Closing rather than rebasing.

rejifald:
Superseded by #503 (production) and #539 (the salvaged guard + docs rule).

PR #529, opened 2026-08-25
feat(marketing): make the «Для української» directory public, and give it a confident voice

rejifald:
Follows #528, which shipped the board layout while the page was still hidden. This makes it findable and rewrites the copy that was written for a page nobody could reach.

## Public

The directory was held back by three separate mechanisms, and the plan is explicit that dropping any one alone leaves a page that is findable but looks unfinished. All three land together:

- `noindex` off both halves.
- Both routes out of `UNLISTED` in `astro.config.mjs`. The set is now empty and the mechanism stays, because the next unfinished page needs it.
- A footer link, with the `localeForUkrainianHref` helper it needed — deliberately absent until now, since an href helper with no caller is dead code.

The fourth item in the plan's checklist, the `MIRRORED_PAGES` entry, was already done; `check:locale-redirects` confirms `/for-ukrainian → /uk/for-ukrainian/` in the pre-commit run.

**The footer column is a judgement call**, and the one thing here worth a second opinion. The directory is not a Movar page — it lists other people's projects and says so — so none of the four footer intents ("get it", "understand it", "trust it", "reach us") actually describes it. It sits under "how it works" because that is in practice where everything readable lives, and because a fifth column would break the four-column row this footer was rebuilt to fit. `footer-links.ts` says so at the entry, and nothing depends on it staying there.

## Voice

The page was written as though it expected to be argued with.

- **«Хто сюди потрапляє» is gone** as redundant beside the disclaimer the page opens with. The scope rules remain the review-time criteria; they are simply no longer published. That cost is real and recorded in the plan — an excluded submitter is now answered in a reply rather than by a standard they can read first. If rejections start needing an explanation more than once, the criteria belong back, on their own route.
- **The verification date is gone entirely.** `lastVerified` stays required in the schema because it is how a human tracks link rot, which was always its real audience. It is not something a reader needs told.
- **Removal-on-request now renders only while the collection names an individual** (`person: true` — currently none). Twenty organisations do not need it announced: anyone who wants their project off writes to us, and saying so unprompted reads as anxious. A named person is different — their entry stands on consent that is theirs to withdraw, so the way to withdraw it has to be visible on the page. Wired to the data rather than left as a note, so it returns on the build that first lists a person.
- **Both remaining promises lost their hedging.** «…приберемо без питань і без публічних пояснень. Прохання від людини виконуємо одразу» → «Ми приберемо будь-кого з переліку на перше ж прохання». «Виправлення так само доречні — якщо ми когось описали неточно, скажіть» → «Помітили неточність — виправимо». Every invitation is one clause, answered in the first person: we add, we fix, we remove.

## Cards and controls

The host moved out of the card's meta row and up under the name, where it reads as the entry's identity. Saying where a link goes before it is clicked is the honest form of a "you are leaving this site" warning — same information, none of the friction, and no JavaScript needed to deliver it, which is why there is no interstitial. It is deliberately **not** a second anchor: it points where the name already points, so linking it would buy a duplicate tab stop and a repeated screen-reader announcement and no way to get anywhere new.

The `20 / 20` counter beside the search box is gone as noise — the chips carry per-sphere counts and the empty state covers nothing-found. It survives as an `sr-only` live region, because a reader who cannot see the grid shrink has no other signal that a search did anything. The two submit actions stack instead of sitting side by side, sharing one width.

## Baselines

All 42 changed marketing baselines were regenerated in the pinned amd64 container. The 38 non-directory ones each grew by **exactly 28px** — one footer row, uniform, so nothing else moved on those pages. The directory itself is 414px shorter: **3,567px against the 5,369px it originally shipped at.**

## One thing this PR cannot do

The plan pairs going public with turning on **"Require approval for all external contributors"** in Settings → Actions → General. That is now urgent rather than pre-emptive: the page is public and invites submissions, the repo is public with forking enabled, and the setting is not reachable through the API. It needs a maintainer click. The changed-paths workflow is the other half of that mitigation and does not substitute for it.


rejifald:
Correction to this PR's description.

It closes by saying "Require approval for all external contributors" is still off and now urgent. **That is wrong — the setting was already on, and at its strictest value.**

`GET /repos/rejifald/movar/actions/permissions/fork-pr-contributor-approval` returns `{"approval_policy":"all_external_contributors"}`, which is exactly that setting. Alongside it, default workflow permissions are `read` with `can_approve_pull_request_reviews: false`, and `directory-submission-guard.yml` triggers on `pull_request` (not `pull_request_target`) and only runs for forks. Both halves of the mitigation the plan asks for are in place.

The error was mine: I checked `actions/permissions/workflow` and `actions/permissions/access`, got a 422 from the latter (it only applies to private repos), concluded the setting was unreachable via API, and inferred it was off. The inference was never verified before being stated.

No action is needed on the repository. The follow-up is to correct the same wrong claim where it landed in `docs/articles/for-ukrainian.plan.md`, which this PR also introduced.

